PR #58, opened 2024-02-05
remove duplicate ci workflow

### LINKED ISSUE #16: ci pipelines

federicazanca:
resolves #16


.github/workflows/ci.yml
```diff
@@ -41,69 +42,44 @@ jobs:
...
     - name: Install python dependencies
-      env:
-        PLUGIN_NAME: aiida-mlip
       run: |
         pip install --upgrade pip
-        pip install -r requirements.txt
-        # Change plugin_name to test we're not specific to "aiida-diff"
-        #cookiecutter --no-input . plugin_name=${PLUGIN_NAME}
-        #pip install -e ${PLUGIN_NAME}[pre-commit,docs,testing]
+        pip install -e .[pre-commit,docs,testing]
```
alinelena, line 81: here I will you 
python -m pip install

alinelena (approved):

.github/workflows/ci.yml
ElliottKasoar: > here I will you python -m pip install

This will likely change assuming we're happy with the poetry changes anyway, but to check, are you just suggesting `pip install` -> `python -m pip install`, rather than getting rid of the square brackets? (I think we need `[pre-commit]` to make sure `pre-commit install` etc. runs, I'm not actually sure we need the other two here)

alinelena: yes I have still traumas after fixing things because pip was the wrong one... so always python -m pip instead of pip alone

ElliottKasoar (changes requested):
As part of "finishing" the CI pipeline for #16, I'm wondering if we ought to do the last step of integration for code coverage. Currently we generate the coverage but don't publish.

I think we probably just need a couple of extra lines (see the [gromacs plugin](https://github.com/PSDI-Biomolecular-team/aiida-gromacs/blob/master/.github/workflows/ci.yml#L61)) assuming we're happy to stick with coveralls, plus some kind of registration on the site and possibly generating a token.

ElliottKasoar:
As discussed with @alinelena, we'll publish code coverage separately - see #60.

Think this should be ready once the `pip`s are replaced.

ElliottKasoar (approved):

oerc0122 (approved):

oerc0122 (approved):

Issue #41, opened 2024-02-02
setup testing infra


alinelena:
> tox with pytest and unit tests.
> 
> be consistent with aiida plugin testing (which is pytest) 

 _Originally posted by @alinelena in [#12](https://github.com/stfc/aiida-mlip/issues/12#issuecomment-1923583539)_

Issue #17, opened 2024-02-01
deployment - pypi


alinelena:

ElliottKasoar:
I think this will be covered by the template action: https://github.com/stfc/aiida-mlip/blob/main/.github/workflows/publish-on-pypi.yml, which should publishe when we tag a version.

As discussed with @alinelena, we may want to publish a latest version as well as stable, tagged versions.

alinelena:
is not really working

ElliottKasoar:
Resolved by https://github.com/stfc/aiida-mlip/commit/600a256a94b8d5279bbf979a783afe709be9fe46.

See: https://pypi.org/project/aiida-mlip/

A few issues remain as described in #71 and #72

PR #59, opened 2024-02-05
Add poetry

### LINKED ISSUE #45: check versions of deps
https://github.com/stfc/aiida-mlip/blob/6df0c26a94078c999415b71316a7717696d304e3/pyproject.toml#L33

the default versions may be a little bit out of date. we shall check them and update them accordingly

ElliottKasoar:
Resolves #45 and #50

Updates pyproject.toml to be compatible with poetry, although I have omitted the .lock file to make this more general.

Notes:
- poetry does not currently conform to [PEP 621](https://peps.python.org/pep-0621/), so the format currently differs from most other modern pyproject.toml files, although support will be implemented [in the next major release](https://github.com/orgs/python-poetry/discussions/5833).
- We cannot use dynamic versioning with this, so now need to specify the version both in pyproject.toml and __init__.py
- Updates will be needed to replace the build/publishing, but these will be in a separate pull request

This also updates the workflows to use poetry installations where possible, bumps `actions` versions in workflows, and bumps `dev` (formerly `testing`) versions in pyproject.toml.

To do:

- [x] Update README with setup instructions

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,16 +1,16 @@
-[build-system]
-# build the package with [flit](https://flit.readthedocs.io)
-requires = ["flit_core >=3.4,<4"]
-build-backend = "flit_core.buildapi"
-
-[project]
-# See https://www.python.org/dev/peps/pep-0621/
+[tool.poetry]
 name = "aiida-mlip"
-dynamic = ["version"]  # read from aiida_mlip/__init__.py
+version = "0.1.0a0"
 description = "machine learning interatomic potentials aiida plugin"
-authors = [{name = "Federica Zanca, Elliott Kasoar, Jacob Wilkins, Alin M. Elena"}]
+authors = [
+    "Federica Zanca",
+    "Elliott Kasoar",
+    "Jacob Wilkins",
+    "Alin M. Elena",
+]
+license = "LICENSE"
 readme = "README.md"
-license = {file = "LICENSE"}
+packages = [{include = "aiida_mlip"}]
 classifiers = [
     "Programming Language :: Python",
     "Intended Audience :: Science/Research",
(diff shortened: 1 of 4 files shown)


ElliottKasoar:
> > * Updates will be needed to replace the build/publishing, but these will be in a separate pull request
> 
> @ElliottKasoar can you open issues for these so we do not loose them once we merge the pr

I think this is already covered by #17 and #18

ElliottKasoar:
This will need rebasing once #58 is merged. I may also drop Python 3.12 back down to 3.11, as discussed with @alinelena, to minimise compatibility issues with [janus](https://github.com/stfc/janus)(-core)

alinelena:
@ElliottKasoar i think some disconnect happened between the ci duplication removal pr #58 and this one... so practically your ci_orig changes need to be ported to ci.yaml... once done is mergeable

alinelena (approved):

alinelena (dismissed):

pyproject.toml
```diff
@@ -99,12 +107,12 @@ known_aiida = ['aiida']
 [tool.tox]
 legacy_tox_ini = """
 [tox]
-envlist = py38
+envlist = py39
```
oerc0122, line 110: Is this the only environment we're wanting? Maybe I'm misreading this as I've not used `tox envlist`

oerc0122 (approved):
Looks fine beyond that clarification for me. Though can you still install this e.g. through `pip` without `poetry`? Because for most people that's probably more convenient.

pyproject.toml
ElliottKasoar: @federicazanca is currently working the tox framework for #41, so the intention here is to do the minimum update to be consistent, rather than the final set of versions.

ElliottKasoar:
> Looks fine beyond that clarification for me. Though can you still install this e.g. through `pip` without `poetry`? Because for most people that's probably more convenient.

Assuming you mean pip installing the repository, `pip install .` will install the main dependencies (under `[tool.poetry.dependencies]`), but I don't think `[docs]` etc. works for the other grouped dependencies, which I suspect is the consequence of the current misalignment with other formats.

oerc0122 (approved):

PR #62, opened 2024-02-06
add doi badge, shall be autopoulated at the first release fixes #52

### LINKED ISSUE #52: add zenodo integration

alinelena:


README.md
```diff
@@ -3,6 +3,7 @@
 [![Docs status][docs-badge]][docs-link]
 [![PyPI version][pypi-badge]][pypi-link]
 [![License][license-badge]][license-link]
+[![DOI](https://zenodo.org/badge/750834002.svg)](https://zenodo.org/badge/latestdoi/750834002)
```
ElliottKasoar, line 6: Doesn't matter too much but potentially could be worth moving the badge/link so together with the others.

ElliottKasoar (changes requested):
I think this has the same problem as add-poetry had from the merge - it's changing ci.yml for some reason, which also means the pre-commit is showing an error.

It might be simpler to reset from main, given that it's basically only a one line change.

PR #65, opened 2024-02-07
Remove unused files

ElliottKasoar:


setup.json
```diff
@@ -1,46 +0,0 @@
-{
```
alinelena, line 1: this is actually needed for the registry is read to populate some info on aiida side

alinelena (changes requested):

setup.json
alinelena: @federicazanca will know that but i think is just following their instructions they expect setup.json to be there

alinelena (approved):

Issue #79, opened 2024-02-13
create data transfer for mlips 


alinelena:
- [x] add the download facility
- [x] create tests
- [x] create documentation

ElliottKasoar:
Resolved by #80

PR #80, opened 2024-02-20
Data transfer #79

federicazanca:
Issue #79 
I created the new aiida data type ModelData for download and saving model files in database (in /aiida_mlip/data/model.py) 
tests in /tests/data/test_model.py
docs in docs/source/data/model.md

The data type allows to save the a file in the aiida database, so data provenance is stored.
It stores info on the file path, model name and architecture, and has implemented a download function to download the model from a link.
I did write something to avoid having the same file twice, I don't know if I did it right.
However the download function is something outside of aiida storage, so it's just a way to help in the process of getting the model file and organising it. The aiida part is just when you save the file you already have as a ModelData (that inherits from SingleFileData which is already an aiida class).

Aiida already gives the possibility to enable caching for specific data types or plugins, so I would just do that to avoid storing the same model in the database twice.

About the tests. At the moment I am downloading a random file from the repo itself to test the download function. It's probably not the ideal thing to do, but also downloading the model everytime you do the test might not be the ideal thing to do? But the point is that I want to check that the download itself works, so I'm not sure what's the usual thing to do in this case.
They work tho so that's good.

I still have to add some test about checking that the thing is actually stored in aiida I guess?


ElliottKasoar:
I need to test this out properly, but looks good so far.

From a stylistic perspective, could you try to add type hints to functions where they're missing, and ideally include similar descriptions in the docstrings. I tend to use a [numpy-style](https://numpydoc.readthedocs.io/en/latest/format.html) e.g.

```python
"""
Description.

Parameters
----------
filename : str
"""
```
which you can check with something like [pydocstyle](https://www.pydocstyle.org/en/stable/) (deprecated) or ruff (which does a lot more too, but is supposed to be good).

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,125 @@
...
+    # Calculate hash
+    sha256 = hashlib.sha256()
+    with open(file, "rb") as f:
+        for data in iter(lambda: f.read(65536), b""):
+            sha256.update(data)
+    file_hash = sha256.hexdigest()
+    return file_hash
+
+
+def check_existing_file(arch_path, file, file_hash, architecture, cache_path):
```
ElliottKasoar, line 22: I'll have a think myself, but is there a way to reduce the number of levels of nesting in this function? The more levels, the more difficult it becomes to follow generally.

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+class ModelData(SinglefileData):
+    """Class to handle models"""
+
+    def __init__(self, file, filename, architecture: str = None, **kwargs):
+        """Add a file to the node, parse it and set the attributes found.
+
+        :param file: absolute path to the file or a filelike object
+        :param filename: specify filename to use (defaults to name of provided file).
```
ElliottKasoar, line 47: It may be more obvious when I actually use this, but I'm not sure it's as clear as it could be what the difference between `file` and `filepath` is.

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,125 @@
...
+
+                if arch_path == cache_path and arch_path != ex_file_path:
+                    architecture = ex_file_path.name
+                    file.unlink()
+                    return existing_file, architecture
+    return file, architecture
+
+
+class ModelData(SinglefileData):
+    """Class to handle models"""
```
ElliottKasoar, line 41: It might be worth making it slightly more clear what model refers to (MLIP?), and very roughly what handling involves (loading/saving?).

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,125 @@
...
+        :param filename: specify filename to use (defaults to name of provided file).
+        """
+
+        super().set_file(file, filename, **kwargs)
+
+        self.base.attributes.set("architecture", architecture)
+
+    @classmethod
+    def local_file(cls, file: str, filename: str = None, architecture: str = None):
+        """Sve a local file as ModelData"""
```
ElliottKasoar, line 67:
```suggestion
        """Save a local file as ModelData."""
```

tests/data/test_model.py
```diff
@@ -0,0 +1,86 @@
...
+    testdir.mkdir(parents=True, exist_ok=False)
+    with open("./tests/data/tmp/mace/test.txt", "w", encoding="utf-8") as f:
+        f.write("file with content\ncontent2\n")
+
+    # Construct a ModelData instance that should use the cached file
+    model = ModelData.download(
+        url="https://raw.githubusercontent.com/stfc/aiida-mlip/main/tests/input_files/file2.txt",
+        cache_dir="./tests/data/tmp/",
+        filename="test_modell.txt",
+        # architecture = "mace"
```
ElliottKasoar, line 71:
```suggestion
```

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,125 @@
...
+
+        # Check if there is already a file named that way and rename it
+        if file.exists():
+            file = file.with_name(file.stem + "_2" + file.suffix)
+            # Check again if the new filename already exists, just in case
+            while file.exists():
+                # If the new filename also exists, keep incrementing the number
+                file = file.with_name(
+                    file.stem[:-2] + str(int(file.stem[-2:]) + 1).zfill(2) + file.suffix
+                )
```
oerc0122, line 106: `with_stem` is Python 3.9+, otherwise can use `with_name` in similar way.
```suggestion
        stem = file.stem
        i = 1
        while file.exists():
            i += 1
            file = file.with_stem(f"{stem}_{i}")
```

tests/data/test_model.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def test_local_file():
+    """Testing that the local file function works"""
+    # Construct a ModelData instance with the local file
+    model = ModelData.local_file("./tests/input_files/model_local_file.txt")
+    # Assert the ModelData contains the content we expect
+    content = model.get_content()
+    with open("./tests/input_files/model_local_file.txt", encoding="utf-8") as f:
+        assert content == f.read()
```
oerc0122, line 16: If this were a `Path`, can use [`read_text`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.read_text)
```suggestion
    assert content == MODEL_FILE.read_text()
```

tests/data/test_model.py
```diff
@@ -0,0 +1,86 @@
...
+    )
+    assert model.architecture == "mace"
+
+
+def test_download_fresh_file():
+    """Test if download works"""
+    # Ensure we do not have the file cached already
+    path_test = Path("./tests/data/tmp/mace/test_download.txt")
+    if path_test.exists():
+        path_test.unlink()
```
oerc0122, line 34: Can do
```suggestion
    path_test.unlink(missing_ok=True)
```

tests/data/test_model.py
```diff
@@ -0,0 +1,86 @@
...
+        filename="test_download.txt",
+        cache_dir="./tests/data/tmp",
+        architecture="mace",
+    )
+
+    # Assert the ModelData contains the content we expect
+    content = model.get_content()
+    assert content == "file with content\ncontent2\n"
+    file_path = Path("./tests/data/tmp/mace/test_download.txt")
+    assert file_path.exists(), f"File {file_path} does not exists."
```
oerc0122, line 48:
```suggestion
    assert file_path.exists(), f"File {file_path} does not exist."
```

aiida_mlip/data/model.py
federicazanca: basically file is the path to the file, while filename is the name you want to give in aiida to save that model.
so if the path to the file is `folder/file1` and you give filename = None, filename will be `file1`. But if you give another name it willuse that to define the file and save it in the database
I called them this way cause this is how they were called in the SinglefileData class, but I also think it's not too intuitive. So I guess I coulkd change it, call them like filepath and modelname or something

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,125 @@
...
+        for data in iter(lambda: f.read(65536), b""):
+            sha256.update(data)
+    file_hash = sha256.hexdigest()
+    return file_hash
+
+
+def check_existing_file(arch_path, file, file_hash, architecture, cache_path):
+    """Function to check if a file already exists and return a file path and architecture"""
+    for existing_file in arch_path.rglob("*"):
+        if existing_file.is_file() and existing_file != file:
```
oerc0122, line 25:
```suggestion
    for existing_file in (ex_file for ex_file in arch_path.rglob("*")
                          if ex_file.is_file() and not ex_file.samefile(file)):
```

tests/data/test_model.py
```diff
@@ -0,0 +1,86 @@
...
+    # Clear test cache
+    shutil.rmtree(Path("./tests/data/tmp/mace/"))
+
+
+def test_no_download_cached_file():
+    """Test if the caching work for avoiding double download"""
+    # Ensure file is not already downloaded
+    cached_file_path = Path("./tests/data/tmp/test_modell.txt")
```
oerc0122, line 56:
```suggestion
    cached_file_path = Path("./tests/data/tmp/test_model.txt")
```

tests/data/test_model.py
```diff
@@ -0,0 +1,86 @@
...
+    if cached_file_path.exists():
+        cached_file_path.unlink()
+
+    # Ensure we have the file cached already
+    testdir = Path("./tests/data/tmp/mace/")
+    testdir.mkdir(parents=True, exist_ok=False)
+    with open("./tests/data/tmp/mace/test.txt", "w", encoding="utf-8") as f:
+        f.write("file with content\ncontent2\n")
```
oerc0122, line 64: With path can do:
```suggestion
    (testdir / "test.txt").write_text("file with content\ncontent2\n", encoding="utf-8")
```

tests/data/test_model.py
```diff
@@ -0,0 +1,86 @@
...
+
+    # Construct a ModelData instance that should use the cached file
+    model = ModelData.download(
+        url="https://raw.githubusercontent.com/stfc/aiida-mlip/main/tests/input_files/file2.txt",
+        cache_dir="./tests/data/tmp/",
+        filename="test_modell.txt",
```
oerc0122, line 70:
```suggestion
        filename="test_model.txt",
```

tests/data/test_model.py
```diff
@@ -0,0 +1,86 @@
...
+        # architecture = "mace"
+    )
+
+    # Assert the ModelData contains the content we expect
+    content = model.get_content()
+    assert content == "file with content\ncontent2\n"
+    file_path = Path("./tests/data/tmp/mace/test_modell.txt")
```
oerc0122, line 77:
```suggestion
    file_path = Path("./tests/data/tmp/mace/test_model.txt")
```

docs/source/data/model.md
```diff
@@ -0,0 +1,41 @@
...
+
+Definces a custom data type called `ModelData` in AiiDA, which is a subclass of the `SinglefileData` type.`ModelData` is used to handle model files and provides functionalities for handling local files and downloading files from URLs.
+Additional features compared to `SinglefileData`:
+ - it can take a non absolute path as an argument (*change or leave? also in this case it has to be a tring but this can also be changed to take a path*)
+ - if given a URL it will download the file, save it in a folder of choice (default = ./cache/mlips, if given an architecture it will create a subfolder with that name), and save the file as AiiDA data type. If the file is downloaded twice it will be canceled if there are duplicates in the same folder, however this is not related with AiiDA caching for its own database. For that, enable the caching for the ModelData class, or the whole aiida-mlip plugin (https://aiida.readthedocs.io/projects/aiida-core/en/latest/topics/provenance/caching.html)
+ - *others to be added? e.g. read architecture from file itself if possible?*
+
+## Functions and Classes
+
+### `calculate_hash(file)`
```
oerc0122, line 11: Should this not be coming from `Sphinx` or similar rather than being handwritten?

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,125 @@
...
+            existing_file_hash = calculate_hash(existing_file)
+            if existing_file_hash == file_hash:
```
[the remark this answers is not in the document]
oerc0122: Also, hashing might be expensive, you might want some other simpler checks first, e.g. file size

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,125 @@
...
+        """Sve a local file as ModelData"""
+        file_path = Path(file).resolve()
+        return cls(file=str(file_path), filename=filename, architecture=architecture)
+
+    @classmethod
+    def download(
+        cls,
+        url: str,
+        filename: str = None,
+        cache_dir: str = None,
```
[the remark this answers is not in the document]
federicazanca: the singlefiledata class takes str and Path, so I guess I'll just do the same. I made it initially to only take strings cause it was easier but I guess it makes more sense to take both right?

aiida_mlip/data/model.py
ElliottKasoar: Ah ok, I think matching names with `SinglefileData` is definitely valid, even if the names aren't perfect - consistency is always useful.

Perhaps if you could just elaborate slightly in the docstring (even just adding something similar to `... object whose contents to copy` as is in `SinglefileData` for `file` might be enough)?

aiida_mlip/data/model.py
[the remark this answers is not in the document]
oerc0122: `Path(Path("."))` is still a Path, so just putting `cache_dir = Path(cache_dir)` at the front should be fine.

tests/data/test_model.py
```diff
@@ -0,0 +1,83 @@
...
+    file_path = Path("./tests/data/tmp/mace/test_model.txt")
+    assert file_path.exists() is False, f"File {file_path} exists but it shouldn't."
+    file_path2 = Path("./tests/data/tmp/test_model.txt")
+    assert file_path2.exists() is False, f"File {file_path2} exists but it shouldn't."
+    file_path3 = Path("./tests/data/tmp/mace/test.txt")
+    assert file_path3.exists() is True, f"File {file_path3} should exist"
+    assert model.architecture == "mace"
+
+    # Clear test cache
+    shutil.rmtree(Path("./tests/data/tmp/"))
```
alinelena, line 83: no way to do this with pathlib?

oerc0122: Nope, afraid not. `for x in Path.walk(top_down=True): x.unlink(); Path.unlink()` but this is somewhat less clear.

alinelena: ok I will let it pass then...

aiida_mlip/data/model.py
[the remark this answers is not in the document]
alinelena: we do not do it all the time so we shall be ok... better safe than sorry

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,159 @@
...
+
+import hashlib
+from pathlib import Path
+from urllib import request
+from urllib.parse import urlparse
+
+from aiida.orm import SinglefileData
+
+
+def _calculate_hash(file: str | Path) -> str:
```
ElliottKasoar, line 11: `|` is incompatible with Python 3.9, unless you also include `from __future__ import annotations`. (We decided against using that in janus-core for now, as it seemed to lead to issues with ASE, but there may be other solutions.)

An alternative is `Union`.

The CI should have picked up on this, but you can also test locally with `tox run -e py39` for Python 3.9 tests, or `tox run -e ALL` for all tests.

docs/source/user_guide/api.rst
```diff
@@ -3,9 +3,9 @@ Model Data Type Documentation
 
 Defines a custom data type called `ModelData` in AiiDA, which is a subclass of the `SinglefileData` type. `ModelData` is used to handle model files and provides functionalities for handling local files and downloading files from URLs.
 Additional features compared to `SinglefileData`:
- - it can take a non absolute path as an argument
- - if given a URL it will download the file, save it in a folder of choice (default = ./cache/mlips, if given an architecture it will create a subfolder with that name), and save the file as AiiDA data type. If the file is downloaded twice it will be canceled if there are duplicates in the same folder, however this is not related with AiiDA caching for its own database. For that, enable the caching for the ModelData class, or the whole aiida-mlip plugin (https://aiida.readthedocs.io/projects/aiida-core/en/latest/topics/provenance/caching.html)
- - *others to be added? e.g. read architecture from file itself if possible?*
+- it can take a non absolute path as an argument
```
oerc0122, line 6: This was the unexpected indentation, but you need a newline here
```suggestion

- it can take a non absolute path as an argument
```

docs/source/user_guide/api.rst
```diff
@@ -14,10 +14,10 @@ Usage
 
 .. code-block:: python
 
-    model = ModelData.local_file('/path/to/file', filename='model', architecture='mace')
+model = ModelData.local_file('/path/to/file', filename='model', architecture='mace')
```
oerc0122, line 17: This is no longer in the code block:
```suggestion
    model = ModelData.local_file('/path/to/file', filename='model', architecture='mace')
```

docs/source/user_guide/api.rst
```diff
@@ -14,10 +14,10 @@ Usage
...
 
 - To download a file and save it as a `ModelData` object:
 
 .. code-block:: python
 
-    model = ModelData.download('http://yoururl.com/model', filename='model', cache_dir='/home/mlip/', architecture='mace')
+model = ModelData.download('http://yoururl.com/model', filename='model', cache_dir='/home/mlip/', architecture='mace')
```
oerc0122, line 23:
```suggestion
    model = ModelData.download('http://yoururl.com/model', filename='model', cache_dir='/home/mlip/', architecture='mace')
```

oerc0122 (changes requested):

docs/source/user_guide/api.rst
```diff
@@ -0,0 +1,23 @@
...
+
+.. code-block:: python
+
+model = ModelData.local_file('/path/to/file', filename='model', architecture='mace')
+
+- To download a file and save it as a `ModelData` object:
+
+.. code-block:: python
+
+model = ModelData.download('http://yoururl.com/model', filename='model', cache_dir='/home/mlip/', architecture='mace')
```
ElliottKasoar, line 23: I'm not sure if it's actually dangerous (and I know it's not a real link) but maybe worth noting that FortiClient thinks this link could be malicious.

oerc0122: You may want to use a [`.test` TLD](https://en.wikipedia.org/wiki/.test)

aiida_mlip/data/model.py
federicazanca: closing this as now it's a bit different anyway following jacob's suggestions, but might be possible to make it better

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,159 @@
...
+    sha256 = hashlib.sha256()
+    with open(file, "rb") as f:
+        # calculating sha in chunks rather than 1 large pass
+        while data := f.read(buf_size):
+            sha256.update(data)
+    file_hash = sha256.hexdigest()
+    return file_hash
+
+
+def _check_existing_file(
```
ElliottKasoar, line 26: Out of interest, is there a benefit to having this function and `_calculate_hash` defined outside of the class?

tests/data/test_model.py
```diff
@@ -0,0 +1,83 @@
...
+        architecture="mace",
+    )
+
+    # Assert the ModelData contains the content we expect
+    content = model.get_content()
+    assert content == "file with content\ncontent2\n"
+    file_path = Path("./tests/data/tmp/mace/test_download.txt")
+    assert file_path.exists(), f"File {file_path} does not exist."
+    # Clear test cache
+    shutil.rmtree(Path("./tests/data/tmp/mace/"))
```
ElliottKasoar, line 49: Would it work to use pytest's [tmpdir](https://docs.pytest.org/en/6.2.x/tmpdir.html) rather than making/deleting our own?

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,159 @@
...
+    param architecture: mlip architecture of the model
+    param cache_path: Path to the parent folder of arch_path (str or pathlib.Path)
+    return: A tuple of a PAth and a str containing the path of the file of interest and the architecture
+    """
+    file_hash = _calculate_hash(file)
+    for existing_file in (
+        ex_file
+        for ex_file in arch_path.rglob("*")
+        if ex_file.is_file() and not ex_file.samefile(file)
+    ):
```
oerc0122, line 41: I think this may be cleaner as a filter
```suggestion
    def is_diff_file(curr_path: Path):
        return curr_path.is_file() and not curr_path.samefile(path)

    for existing_file in filter(is_diff_file, arch_path.rglob("*")):
```

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,159 @@
...
+    with open(file, "rb") as f:
+        # calculating sha in chunks rather than 1 large pass
+        while data := f.read(buf_size):
+            sha256.update(data)
+    file_hash = sha256.hexdigest()
+    return file_hash
+
+
+def _check_existing_file(
+    arch_path: str | Path, file: str | Path, architecture: str, cache_path: str | Path
```
oerc0122, line 27:
```suggestion
    arch_path: str | Path, file: str | Path, architecture: str, cache_path: str | Path
```
str | Path can be `os.PathLike`

tests/data/test_model.py
```diff
@@ -0,0 +1,83 @@
...
+from pathlib import Path
+import shutil
+
+from aiida_mlip.data.model import ModelData
+
+
+def test_local_file():
+    """Testing that the local file function works"""
+    # Construct a ModelData instance with the local file
+    model_path = Path("./tests/input_files/model_local_file.txt")
```
oerc0122, line 12: May not want to start from `.` if run from elsewhere. `Path(__file__).parent` references the directory which contains this file

tests/data/test_model.py
oerc0122: Or Python stdlib's [tempfile](https://docs.python.org/3/library/tempfile.html)

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,159 @@
...
+        """Save a local file as ModelData
+        :param file: path to the file (str or pathlib.Path)
+        :param filename: specify filename to use (defaults to name of provided file)
+        :param architecture: specify architecture in which the model is used (default = None)
+        """
+        file_path = Path(file).resolve()
+        return cls(file=file_path, filename=filename, architecture=architecture)
+
+    @classmethod
+    # pylint: disable=R0913
```
ElliottKasoar, line 100: `too-many-arguments` is probably preferable to the code, so it's clear what you're disabling.

It might also be worth adding a value for `max-args = ` under `[tool.pylint.format]` in pyproject.toml if this is needed more often. I prefer more having the option of more than five generally, as long as it's something to still consider.

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,243 @@
...
+            Name to be used for the file (defaults to the name of provided file).
+        architecture : Optional[str], optional
+            Architecture information.
+
+        Other Parameters
+        ----------------
+        kwargs : Any
+            Additional keyword arguments.
+        """
+
```
ElliottKasoar, line 107:
```suggestion
```

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,243 @@
...
+            Name to be used for the file (defaults to the name of provided file).
+        architecture : Optional[str], optional
+            Architecture.
+
+        Other Parameters
+        ----------------
+        kwargs : Any
+            Additional keyword arguments.
+        """
+
```
ElliottKasoar, line 134:
```suggestion
```

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,243 @@
...
+            Name to be used for the file (defaults to the name of provided file).
+        architecture : Optional[str], optional
+            Architecture.
+
+        Returns
+        -------
+        ModelData
+            A ModelData instance.
+        """
+
```
ElliottKasoar, line 162:
```suggestion
```

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,243 @@
...
+            Architecture.
+        force_download : Optional[bool], optional
+            True to keep the downloaded model even if there are duplicates (default: False).
+
+        Returns
+        -------
+        ModelData
+            A ModelData instance.
+        """
+
```
ElliottKasoar, line 196:
```suggestion
```

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,243 @@
...
+    @property
+    def architecture(self) -> str:
+        """Return the architecture.
+
+        Returns
+        -------
+        str
+            Architecture.
+        """
+
```
ElliottKasoar, line 242:
```suggestion
```

tests/data/test_model.py
```diff
@@ -0,0 +1,81 @@
...
+        architecture="mace",
+    )
+    assert model.architecture == "mace"
+
+
+def test_download_fresh_file(tmp_path):
+    """Test if download works"""
+    # Ensure we do not have the file cached already
+    path_test = tmp_path / "mace" / "test_download.txt"
+    # path_test = Path("./tests/data/tmp/mace/test_download.txt")
```
ElliottKasoar, line 33:
```suggestion
```

tests/data/test_model.py
```diff
@@ -0,0 +1,81 @@
...
+        filename="test_download.txt",
+        cache_dir=tmp_path,
+        architecture="mace",
+    )
+
+    # Assert the ModelData contains the content we expect
+    content = model.get_content()
+    assert content == "file with content\ncontent2\n"
+    file_path = tmp_path / "mace" / "test_download.txt"
+    # file_path = Path("./tests/data/tmp/mace/test_download.txt")
```
ElliottKasoar, line 48:
```suggestion
```

tests/data/test_model.py
```diff
@@ -0,0 +1,81 @@
...
+    assert file_path.exists(), f"File {file_path} does not exist."
+
+
+def test_no_download_cached_file(tmp_path):
+    """Test if the caching work for avoiding double download"""
+    # Ensure file is not already downloaded
+
```
ElliottKasoar, line 55:
```suggestion
```

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,243 @@
...
+    file: Union[str, Path],
+    cache_path: Union[str, Path],
+    architecture: Optional[str] = None,
+) -> tuple[Path, str]:
+    """Check if a file already exists and return its path and architecture.
+
+    Parameters
+    ----------
+    arch_path : Union[str, Path]
+        Path to the folder containing the file.
```
ElliottKasoar, line 47: Do we need to pass both the file and the folder containing the file?

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,243 @@
...
+    file : Union[str, Path]
+        Path to the file.
+    cache_path : Union[str, Path]
+        Path to the parent folder of arch_path.
```
ElliottKasoar, line 51: Similar to above

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,243 @@
...
+        return curr_path.is_file() and not curr_path.samefile(file)
+
+    for existing_file in filter(is_diff_file, arch_path.rglob("*")):
+        if _calculate_hash(existing_file) == file_hash:
+            ex_file_path = Path(existing_file).parent
+            if arch_path == ex_file_path:
+                file.unlink()
+                return existing_file, architecture
+
+            if arch_path == cache_path and arch_path != ex_file_path:
```
ElliottKasoar, line 72:
```suggestion
            if arch_path == cache_path
```
If `arch_path == ex_file_path`, wouldn't the return above be triggered?

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,243 @@
...
+        """Create a ModelData instance from a local file.
+
+        Parameters
+        ----------
+        file : Union[str, Path]
+            Path to the file.
+        filename : Optional[str], optional
+            Name to be used for the file (defaults to the name of provided file).
+        architecture : Optional[str], optional
+            Architecture.
```
ElliottKasoar, line 155: Could be slightly more descriptive?

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,243 @@
...
+        return cls.local_file(file=filepath, architecture=architecture)
+
+    @property
+    def architecture(self) -> str:
+        """Return the architecture.
+
+        Returns
+        -------
+        str
+            Architecture.
```
ElliottKasoar, line 240: Could be slightly more descriptive?

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,243 @@
...
+        """Initialize the ModelData object.
+
+        Parameters
+        ----------
+        file : Union[str, Path]
+            Absolute path to the file.
+        filename : Optional[str], optional
+            Name to be used for the file (defaults to the name of provided file).
+        architecture : Optional[str], optional
+            Architecture information.
```
ElliottKasoar, line 100: It's probably not the easiest to describe, but "information" is quite ambiguous to me

ElliottKasoar (changes requested):

aiida_mlip/data/model.py
ElliottKasoar: `str` isn't `PathLike`, is it?

```python
In [9]: isinstance("./tests", os.PathLike)
Out[9]: False

In [10]: isinstance(pathlib.Path("./tests"), os.PathLike)
Out[10]: True
```

oerc0122: I suppose not, huh. I have always used it as such to indicate to users that the thing should look like a path. Guess I've been using it wrong.

aiida_mlip/data/model.py
federicazanca: prob not, it was easier for me to remember what I'm doing but should change it

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,239 @@
...
+        return curr_path.is_file() and not curr_path.samefile(file)
+
+    for existing_file in filter(is_diff_file, arch_path.rglob("*")):
+        if _calculate_hash(existing_file) == file_hash:
+            ex_file_path = Path(existing_file).parent
+            if arch_path == ex_file_path:
+                file.unlink()
+                return existing_file, architecture
+
+            if arch_path == cache_path
```
ElliottKasoar, line 72:
```suggestion
            if arch_path == cache_path:
```
Sorry, I think my suggested commit removed this

docs/source/user_guide/api.rst
```diff
@@ -0,0 +1,31 @@
...
+Defines a custom data type called `ModelData` in AiiDA, which is a subclass of the `SinglefileData` type. `ModelData` is used to handle model files and provides functionalities for handling local files and downloading files from URLs.
+Additional features compared to `SinglefileData`:
+
+- it can take a relative path as an argument
+
+- it takes the argument "architecture" which is specifically related to the mlip model and it is added to the node attributes.
+
+- if given a URL it will download the file, save it in a folder of choice (default = ./cache/mlips, if given an architecture it will create a subfolder with that name), and save the file as AiiDA data type.  If the file is downloaded twice it will be canceled if there are duplicates in the same folder (unless specified diffeerently with the keyword "force_download=True"). However this is not related with AiiDA caching for its own database. For that, enable the caching for the ModelData class, or the whole aiida-mlip plugin (https://aiida.readthedocs.io/projects/aiida-core/en/latest/topics/provenance/caching.html)
+
+- *other features to be added possibly*
```
ElliottKasoar, line 16:
```suggestion
```

docs/source/user_guide/api.rst
```diff
@@ -0,0 +1,31 @@
...
+ModelData
+---------
+Defines a custom data type called `ModelData` in AiiDA, which is a subclass of the `SinglefileData` type. `ModelData` is used to handle model files and provides functionalities for handling local files and downloading files from URLs.
+Additional features compared to `SinglefileData`:
+
+- it can take a relative path as an argument
+
+- it takes the argument "architecture" which is specifically related to the mlip model and it is added to the node attributes.
+
+- if given a URL it will download the file, save it in a folder of choice (default = ./cache/mlips, if given an architecture it will create a subfolder with that name), and save the file as AiiDA data type.  If the file is downloaded twice it will be canceled if there are duplicates in the same folder (unless specified diffeerently with the keyword "force_download=True"). However this is not related with AiiDA caching for its own database. For that, enable the caching for the ModelData class, or the whole aiida-mlip plugin (https://aiida.readthedocs.io/projects/aiida-core/en/latest/topics/provenance/caching.html)
```
ElliottKasoar, line 14: It might be nice to break this up, as it's quite a big block of text at the moment. Probably just sub-bullets for the elaboration would be enough?

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,224 @@
...
+            # calculating sha in chunks rather than 1 large pass
+            while data := f.read(buf_size):
+                sha256.update(data)
+        file_hash = sha256.hexdigest()
+        return file_hash
+
+    @staticmethod
+    def _check_existing_file(
+        file: Union[str, Path],
+    ) -> Path:
```
ElliottKasoar, line 45:
```suggestion
    def _check_existing_file(file: Union[str, Path]) -> Path:
```

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,224 @@
...
+            The path of the model file of interest (same as input path if no duplicates were found).
+        """
+        file_hash = ModelData._calculate_hash(file)
+
+        def is_diff_file(curr_path: Path):
+            return curr_path.is_file() and not curr_path.samefile(file)
+
+        file_folder = Path(file).parent
+        for existing_file in filter(is_diff_file, file_folder.rglob("*")):
+            if ModelData._calculate_hash(existing_file) == file_hash:
```
ElliottKasoar, line 65: Would it make sense to make this method into a class method and call this via `cls._calculate_hash`?

My feeling is converting it would be slightly preferable, but I'm not sure if it makes a difference.

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,224 @@
...
+                ex_file_folder = Path(existing_file).parent
+                if file_folder == ex_file_folder:
+                    file.unlink()
+                    return Path(existing_file)
+        return Path(file)
```
ElliottKasoar, line 70: I may be misreading, but am I correct in thinking that if you download `/cache/mace/file`, this will search for existing files matching `file` in  `/cache/` recursively?

Then, if the file already exists in `/cache/mace/`, it discards the new file, but are we not ignoring matches in any other directory (say, `/cache/mace_mp/file`). That would be picked up as `existing_file` (I think?), but that information is unused?

Maybe we should throw a warning or something (if we still want to check all other architecture directories, which doesn't seem unreasonable)?

ElliottKasoar (changes requested):

aiida_mlip/data/model.py
federicazanca: So, initially, the idea was that if you have `architecture` it will only check in `cache_dir/architecture` (so if there is the same model in `cache_dir/` it will not catch it), while if you don't have `architecture` it will check in all the subfolders of `cache_dir`, so if it finds a duplicate in a subfolder, it automatically knows what architecture it is (and this is the only reason that recursive check made sense).
Now that I made `architecture` mandatory, I would expect a folder `cache_dir` to not have any files but just subfolders with the names of different architectures. So basically, the check will happen inside the single subfolder, not in `cache_dir`.

So yes, if there are duplicates outside the folder that has been checked, we will miss them, but in theory there should not be (cause I am assuming that a specific model only works for a specific architecture, which I'm not 100% sure it's true, and that every file in the cache_dir has an architecture already assigned, so it's already in the right folder, and that everything in cache_dir is created with the ModelData class, which is also not necessarily true, cause one could  add things to it manually... but in that case that's their fault?)

ElliottKasoar: Do we still need `if file_folder == ex_file_folder` then if we're only checking the same directory?

I don't think we necessarily need to worry about what people could do manually, but I can imagine slightly different names for the same model, but maybe that's a problem for the user.

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,220 @@
...
+        """
+        file_hash = cls._calculate_hash(file)
+
+        def is_diff_file(curr_path: Path):
+            return curr_path.is_file() and not curr_path.samefile(file)
+
+        file_folder = Path(file).parent
+        for existing_file in filter(is_diff_file, file_folder.rglob("*")):
+            if cls._calculate_hash(existing_file) == file_hash:
+                file.unlink()
```
alinelena, line 64: shall we use a try block in here in case the delete fails?

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,220 @@
...
+        arch_dir = cache_dir / architecture if architecture else cache_dir
+
+        # cache_path = cache_dir.resolve()
+        arch_path = arch_dir.resolve()
+        arch_path.mkdir(parents=True, exist_ok=True)
+
+        model_name = urlparse(url).path.split("/")[-1]
+
+        file = arch_path
+        if filename:
```
alinelena, line 187: use the one line if as above

tests/data/test_model.py
```diff
@@ -0,0 +1,79 @@
...
+    assert content == "file with content\ncontent2\n"
+    file_path = tmp_path / "mace" / "test_download.txt"
+    assert file_path.exists(), f"File {file_path} does not exist."
+
+
+def test_no_download_cached_file(tmp_path):
+    """Test if the caching work for avoiding double download"""
+    # Ensure file is not already downloaded
+    cached_file_path = tmp_path / "test_model.txt"
+    if cached_file_path.exists():
```
alinelena, line 54: I will remove the if and pass to unlink(missing_ok=True)

aiida_mlip/data/model.py
alinelena: no need ignore me

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,220 @@
...
+        file_hash = cls._calculate_hash(file)
+
+        def is_diff_file(curr_path: Path):
+            return curr_path.is_file() and not curr_path.samefile(file)
+
+        file_folder = Path(file).parent
+        for existing_file in filter(is_diff_file, file_folder.rglob("*")):
+            if cls._calculate_hash(existing_file) == file_hash:
+                file.unlink()
+                return Path(existing_file)
```
oerc0122, line 65: Existing file should already be a `Path` here as it's coming from glob. Double-check.

alinelena (approved):

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,220 @@
...
+        force_download : Optional[bool], optional
+            True to keep the downloaded model even if there are duplicates (default: False).
+
+        Returns
+        -------
+        ModelData
+            A ModelData instance.
+        """
+        cache_dir = Path(cache_dir if cache_dir else "~/.cache/mlips/")
+        arch_dir = cache_dir / architecture if architecture else cache_dir
```
oerc0122, line 178: While I know the result due to precedence, at first glance this seems like it might create `cache_dir / cache_dir`
```suggestion
        arch_dir = (cache_dir / architecture) if architecture else cache_dir
```

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,220 @@
...
+
+        model_name = urlparse(url).path.split("/")[-1]
+
+        file = arch_path
+        if filename:
+            file /= filename
+        else:
+            file /= model_name
+
+        # Check if there is already a file named that way and rename it
```
oerc0122, line 188:
```suggestion
        # If file already exists, use next indexed name
```

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,220 @@
...
+
+        Returns
+        -------
+        ModelData
+            A ModelData instance.
+        """
+        file_path = Path(file).resolve()
+        return cls(file=file_path, architecture=architecture, filename=filename)
+
+    @classmethod
```
oerc0122, line 147: I'm curious why this and `check_existing_file` are `classmethod`s when they seem to depend only on aspects of self.

alinelena (dismissed):

oerc0122 (changes requested):

oerc0122 (approved):

alinelena (approved):

ElliottKasoar (dismissed):

aiida_mlip/data/model.py
```diff
@@ -0,0 +1,216 @@
...
+        sha256 = hashlib.sha256()
+        with open(file, "rb") as f:
+            # calculating sha in chunks rather than 1 large pass
+            while data := f.read(buf_size):
+                sha256.update(data)
+        file_hash = sha256.hexdigest()
+        return file_hash
+
+    @classmethod
+    def _check_existing_file(cls, file: Union[str, Path]) -> Path:
```
alinelena, line 43: now that I am looking one more time, what does it happen if you pass a relative path?

tests/data/test_model.py
```diff
@@ -0,0 +1,78 @@
...
+    # Assert the ModelData contains the content we expect
+    content = model.get_content()
+    assert content == "file with content\ncontent2\n"
+    file_path = tmp_path / "mace" / "test_model.txt"
+    assert not file_path.exists(), f"File {file_path} exists but it shouldn't."
+    file_path2 = tmp_path / "test_model.txt"
+    assert not file_path2.exists(), f"File {file_path2} exists but it shouldn't."
+    file_path3 = tmp_path / "mace" / "test.txt"
+    assert file_path3.exists(), f"File {file_path3} should exist"
+    assert model.architecture == "mace"
```
alinelena, line 78: add a test for local file but relative path...

alinelena (approved):

ElliottKasoar (approved):

oerc0122 (approved):

PR #84, opened 2024-03-01
single point calcjob 

federicazanca:
implementing calcjob in aiida for single point calculation
work in progress


aiida_mlip/calculations/singlepoint.py
```diff
@@ -0,0 +1,107 @@
...
+    def define(cls, spec: CalcJobProcessSpec):
+        """Define the process specification, including its inputs, outputs and known exit codes.
+
+        :param spec: the calculation job process spec to define.
+        """
+        super().define(spec)
+
+        # Define inputs
+        spec.input("model", valid_type=ModelData)
+        spec.input("structure", valid_type=StructureData, help="The inputs structure.")
```
alinelena, line 26: the input structure

pyproject.toml
```diff
@@ -29,6 +29,7 @@ documentation = "https://stfc.github.io/aiida-mlip/"
 python = "^3.9"
 aiida-core = "^2.5"
 voluptuous = "^0.14"
+PyCifRW = "^4.4.6"
```
ElliottKasoar, line 32:
```suggestion
ase = "^3.22.1"
PyCifRW = "^4.4.6"
voluptuous = "^0.14"
```
I think ASE needs to be added as dependency somewhere to stop the pre-commit failing.

I possibly wouldn't actually commit this suggestion currently, since it's only actually used for the example, but it'd be needed if you use ASE in the main code.

ElliottKasoar: Oh not sure if it was before, but it's needed in `aiida_mlip/parsers.py`, so you probably do need to add it as a dependency?

aiida_mlip/calculations/singlepoint.py
```diff
@@ -0,0 +1,210 @@
...
+
+from aiida_mlip.data.model import ModelData
+
+
+class Singlepoint(CalcJob):
+    "Calcjob implementation to run single point calculations using mlips"
+
+    _DEFAULT_INPUT_FILE = "aiida.cif"
+    _DEFAULT_OUTPUT_FILE = "aiida.log"
+    _XYZ_OUTPUT = Str("aiida.xyz")
```
alinelena, line 18: this shall be aiida-results.xyz

aiida_mlip/calculations/singlepoint.py
```diff
@@ -0,0 +1,226 @@
...
+            The inputs dictionary.
+
+        Returns
+        -------
+        str or None
+            Error message if validation fails, None otherwise.
+        """
+
+        # Check if  structure is given as it is required
+        for key in "structure":
```
oerc0122, line 136: Does this want to be a string?
```suggestion
for key in structure:
```

aiida_mlip/calculations/singlepoint.py
```diff
@@ -0,0 +1,226 @@
...
+            if key not in value:
+                return f"required value was not provided for the `{key}` namespace."
```
oerc0122, line 138:
```suggestion
        if any(key not in value for key in structure):
            return f"required value was not provided for the `{key}` namespace."
```

aiida_mlip/calculations/singlepoint.py
```diff
@@ -0,0 +1,226 @@
...
+        architecture = (
+            str((self.inputs.model).architecture)
+            if self.inputs.model
+            else str(self.inputs.architecture.value)
+        )
+        model_path = (
+            str((self.inputs.model).filepath)
+            if self.inputs.model
+            else ModelData.download(
+                "http://tinyurl.com/46jrkm3v", architecture
```
oerc0122, line 170: I do not like seeing a random tinyurl link downloaded automatically. This is not safe.

aiida_mlip/calculations/singlepoint.py
```diff
@@ -0,0 +1,226 @@
...
+        # Initialize cmdline_params as an empty list
+        codeinfo.cmdline_params = []
+        # Adding command line params for when we run janus
+        codeinfo.cmdline_params.append(calctype)
+        for flag, value in cmd_line.items():
+            if isinstance(value, dict):
+                # If the value is a dictionary, format it properly
+                formatted_value = (
+                    "{" + ", ".join([f"'{k}': '{v}'" for k, v in value.items()]) + "}"
+                )
```
oerc0122, line 206: This is just a `repr` of the `dict`?

oerc0122: `join` works with generators
```suggestion
                formatted_value = (
                    "{" + ", ".join(f"'{k}': '{v}'" for k, v in value.items()) + "}"
                )
```

aiida_mlip/calculations/singlepoint.py
```diff
@@ -0,0 +1,226 @@
...
+
+        calcinfo = datastructures.CalcInfo()
+        calcinfo.codes_info = [codeinfo]
+        # Save the info about the node where the calc is stored
+        calcinfo.uuid = str(self.uuid)
+        # Retrieve output files
+        calcinfo.retrieve_list = []
+        calcinfo.retrieve_list.append(self.metadata.options.output_filename)
+        calcinfo.retrieve_list.append(xyz_filename)
+        calcinfo.retrieve_list.append(self.uuid)
```
oerc0122, line 224:
```suggestion
        calcinfo.retrieve_list = [self.metadata.options.output_filename,
                                  xyz_filename,
                                  self.uuid]
```

aiida_mlip/parsers.py
```diff
@@ -4,16 +4,40 @@
...
+from aiida.orm import Dict, SinglefileData
 from aiida.parsers.parser import Parser
 from aiida.plugins import CalculationFactory
 
-DiffCalculation = CalculationFactory("mlip")
+Singlepointcalc = CalculationFactory("janus.sp")
+
+
+def convert_numpy(dictionary: dict) -> dict:
+    """A function to convert numpy arrays in lists as the nodes won't store the otherwise
```
oerc0122, line 22:
```suggestion
    """A function to convert numpy `ndarrays` in `dictionary` into `list`s as the nodes won't store them otherwise
```

aiida_mlip/parsers.py
```diff
@@ -22,37 +46,66 @@ def __init__(self, node):
...
     def parse(self, **kwargs):
         """
-        Parse outputs, store results in database.
+        Parse outputs, store results in the database.
 
-        :returns: an exit code, if parsing fails (or nothing if parsing succeeds)
+        Returns
+        -------
+        int
+            An exit code
```
oerc0122, line 67: What exit codes?

examples/calculations/submit_singlepoint.py
```diff
@@ -0,0 +1,177 @@
...
+    """
+    if string is None:
+        model = None
+        return model
+    file_path = Path(string)
+    print(file_path)
+    if file_path.is_file():
+        model = ModelData.local_file(file_path, architecture=architecture)
+        return model
+    model = ModelData.download(string, architecture=architecture)
```
oerc0122, line 48: Either want single return or no intermediate variables, personally, I prefer single return. 

**N.B.** for other style replace all `model = ` here with `return`
```suggestion
    if string is None:
        model = None
    elif (file_path := Path(string)).is_file():
        print(file_path)
        model = ModelData.local_file(file_path, architecture=architecture)
    else:
        model = ModelData.download(string, architecture=architecture)
```

examples/calculations/submit_singlepoint.py
```diff
@@ -0,0 +1,177 @@
...
+        return structure
+    except ValueError as exc:
+        # If conversion to int fails, it should be a string
+        if Path.exists(value):
+            structure = StructureData(ase=read(value))
+            return structure
+        # If it's none, return error
+        raise click.BadParameter(
+            f"Invalid input: {value}. Must be either node PK (int) or a valid path to a structure file."
+        ) from exc
```
oerc0122, line 92:
```suggestion
    if value is None:
        structure = StructureData(ase=bulk("NaCl", "rocksalt", 5.63))
    elif isinstance(value, int) or (isinstance(value, str) and value.isdigit()):
        structure_pk = int(value)
        structure = load_node(structure_pk)
    elif Path.exists(value):
        structure = StructureData(ase=read(value))
    else:
        raise click.BadParameter(
            f"Invalid input: {value}. Must be either node PK (int) or a valid path to a structure file."
        )
    return structure
```

examples/calculations/submit_singlepoint.py
```diff
@@ -0,0 +1,177 @@
...
+        "calctype": Str(params["calctype"]),
+        "model": model,
+        "precision": Str(params["precision"]),
+        "device": Str(params["device"]),
+    }
+
+    # Submit calculation to aiida
+    r, s = run_get_node(Singlepointcalc, **inputs)
+    print(r)
+    print(s)
```
oerc0122, line 133: Why `r` and `s` here? Also this function returns `None`

examples/calculations/submit_singlepoint.py
```diff
@@ -0,0 +1,177 @@
...
+    file_path = Path(string)
+    print(file_path)
+    if file_path.is_file():
+        model = ModelData.local_file(file_path, architecture=architecture)
+        return model
+    model = ModelData.download(string, architecture=architecture)
+    return model
+
+
+def load_structure(value: Union[str, int, None]):
```
oerc0122, line 52: Type-hints on return (and on other functions)

oerc0122 (changes requested):

aiida_mlip/calculations/singlepoint.py
federicazanca: fair enough, changed with the model in janus-core

aiida_mlip/parsers.py
alinelena: i think you missed the return in the if above

aiida_mlip/calculations/singlepoint.py
```diff
@@ -0,0 +1,226 @@
...
+        ----------
+        spec : aiida.engine.CalcJobProcessSpec
+            the calculation job process spec to define.
+        """
+        super().define(spec)
+
+        # Define inputs
+        spec.input(
+            "calctype",
+            valid_type=Str,
```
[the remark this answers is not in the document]
federicazanca: I don't know what does not work in what I'm doing, but at the moment I just fixed the validate_inputs class to do this check. Cause using the validator this way I was getting errors

examples/calculations/submit_singlepoint.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+# Arguments and options to give to the cli when running the script
+@click.command("cli")
+@click.argument("codelabel", type=str)
+@click.option("--calctype", default="singlepoint", type=str)
+@click.option(
+    "--file",
+    default=None,
+    type=Union[str, int],
```
ElliottKasoar, line 134: Can/should this be a [Path](https://click.palletsprojects.com/en/8.1.x/arguments/#file-path-arguments) rather than a string? (This also applies to other path inputs)

examples/calculations/submit_singlepoint.py
```diff
@@ -0,0 +1,168 @@
...
+
+from aiida.common import NotExistent
+from aiida.engine import run_get_node
+from aiida.orm import Code, Str, StructureData, load_node
+from aiida.plugins import CalculationFactory
+
+from aiida_mlip.data.model import ModelData
+
+# StructureData = DataFactory("core.structure")
+
```
ElliottKasoar, line 19:
```suggestion
```

examples/calculations/submit_singlepoint.py
```diff
@@ -0,0 +1,168 @@
...
+
+def load_model(string: Union[str, None], architecture: str) -> ModelData:
```
ElliottKasoar, line 21: Can `string` be a `Path`?

Something like `model` or `model_path` may also be clearer?

examples/calculations/submit_singlepoint.py
```diff
@@ -0,0 +1,168 @@
...
+    if string is None:
+        model = None
+    elif (file_path := Path(string)).is_file():
+        model = ModelData.local_file(file_path, architecture=architecture)
+    else:
+        model = ModelData.download(string, architecture=architecture)
+    return model
+
+
+def load_structure(value: Union[str, int, None]) -> StructureData:
```
ElliottKasoar, line 49: Similar to above, can this be a `Path`, and `value` isn't a particularly clear name. `struct_name` or `struct_path` maybe?

examples/calculations/submit_singlepoint.py
```diff
@@ -0,0 +1,168 @@
...
+    click.BadParameter
+        If the input is not a valid path to a structure file or a node PK.
+    """
+    if value is None:
+        structure = StructureData(ase=bulk("NaCl", "rocksalt", 5.63))
+    elif isinstance(value, int) or (isinstance(value, str) and value.isdigit()):
+        structure_pk = int(value)
+        structure = load_node(structure_pk)
+    elif Path.exists(value):
+        structure = StructureData(ase=read(value))
```
ElliottKasoar, line 79: Does this have to be a single structure rather than a trajectory?

examples/calculations/submit_singlepoint.py
```diff
@@ -0,0 +1,168 @@
...
+    tuple
+        A tuple containing the exit code and the node of the calculation.
+    """
+    structure = load_structure(params["file"])
+
+    # Select model to use
+    model = load_model(params["model"], params["architecture"])
+
+    # Select calculation to use
+    Singlepointcalc = CalculationFactory("janus.sp")
```
ElliottKasoar, line 107: Probably ought to be camel case (`SinglepointCalc`) here and elsewhere?

ElliottKasoar (changes requested):

alinelena:
> * [ ]  Understand what's the deal with the file in the test, I am looking into this(ame)

the issue is this 
image

one of the transformation adds empty spaces...

alinelena:
ok solution is to use the file

[a.zip](https://github.com/stfc/aiida-mlip/files/14578346/a.zip)

we shall understand why the structure written by aiida has that empty spaces.

also this need to be checked

```
  /home/drFaustroll/playground/mlip/aiida-mlip-federica/.tox/py311/lib/python3.11/site-packages/aiida/orm/nodes/data/code/legacy.py:42: AiidaDeprecationWarning: The `Code` class is deprecated. To create an instance, use the `aiida.orm.nodes.data.code.installed.InstalledCode` or `aiida.orm.nodes.data.code.portable.PortableCode` for a "remote" or "local" code, respectively. If you are using this class to compare type, e.g. in `isinstance`, use `aiida.orm.nodes.data.code.abstract.AbstractCode`. (this will be removed in v3)
    warn_deprecation(
```

pyproject.toml
```diff
@@ -62,16 +64,16 @@ build-backend = "poetry.core.masonry.api"
...
+"janus.sp" = "aiida_mlip.calculations.singlepoint:Singlepoint"
 
 [tool.poetry.plugins."aiida.parsers"]
-"mlip" = "aiida_mlip.parsers:DiffParser"
+"janus.parser" = "aiida_mlip.parsers:SPParser"
 
-[tool.poetry.plugins."aiida.cmdline.data"]
-"mlip" = "aiida_mlip.cli:data_cli"
+#[tool.poetry.plugins."aiida.cmdline.data"]
+#"mlip" = "aiida_mlip.cli:data_cli"
```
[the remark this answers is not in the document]
federicazanca: can I leave it for the future when I need it as a reminder or is it bad practice?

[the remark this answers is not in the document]
ElliottKasoar: If you know something along these lines needs adding later, ideally I think I'd open an issue and put this code there as a reference for how to approach it.

It makes it more visible if someone else needs to work on it/you forget.

federicazanca:
new to do list
- [ ] make tests work in the CI. For setting up the code in aiida I need the executable path. How do I get that for janus? I'm trying to copy from quantum espresso and other examples but can't make it work 
- [ ] docs vs pre-commit. for the singlepoint class, to create the docs it wants me to add *args and **kwargs to the docstring, but if I do it makes the pre-commits fail caus it does not recognise the parameters. I checked in the conf.py and pyproject file if there are differences in the setup but might have missed them
- [ ] fix aiida deprecation warnings (should be easy)
- [ ] do your corrections

I accept suggestions for the first 2 as I am clueless

ElliottKasoar:
The issue with the janus executable is probably because it's pip installing, but the version on pypi is pre-CLI. (There may be other issues, but I assume that's part of it)

alinelena:
> new to do list
> 
> * [ ]  make tests work in the CI. For setting up the code in aiida I need the executable path. How do I get that for janus? I'm trying to copy from quantum espresso and other examples but can't make it work
tagged a new version... so if you rerun the ci that bit shall be fixed.

tests/conftest.py
```diff
@@ -0,0 +1,114 @@
...
+    """Return a localhost `Computer`."""
+    localhost = aiida_localhost
+    localhost.set_default_mpiprocs_per_machine(1)
+    return localhost
+
+
+@pytest.fixture(scope="function")
+def janus_code(aiida_local_code_factory):
+    """Get the janus code."""
+    janus_path = os.environ.get("JANUS_PATH")
```
alinelena, line 62: this does not get the path as expected... 
if you set janus_path = "janus"

alll will work but in theory env variable shall work correctly

.github/workflows/ci.yml
```diff
@@ -48,10 +48,15 @@ jobs:
         poetry install --with dev
```
ElliottKasoar, line 48:
```suggestion
        poetry install --with dev
        echo "JANUS_PATH=$(poetry run which janus)" >> $GITHUB_ENV
```

.github/workflows/ci.yml
```diff
@@ -48,10 +48,15 @@ jobs:
...
       env:
-        # show timings of tests
+        # show timings of test
         PYTEST_ADDOPTS: "--durations=0"
-      run: poetry run pytest --cov aiida_mlip --cov-append .
+        JANUS_PATH: $(which janus)
+      run: |
+        pip list
+        which janus
+        poetry run pytest --cov aiida_mlip --cov-append .
```
ElliottKasoar, line 59:
```suggestion
        JANUS_PATH: ${{ env.JANUS_PATH }}
      run: |
        poetry run pytest --cov aiida_mlip --cov-append .
```

ElliottKasoar (changes requested):

.github/workflows/ci.yml
```diff
@@ -46,12 +46,16 @@ jobs:
...
     - name: Run test suite
+
       env:
-        # show timings of tests
+        # show timings of test
         PYTEST_ADDOPTS: "--durations=0"
-      run: poetry run pytest --cov aiida_mlip --cov-append .
+        JANUS_PATH: ${{ env.JANUS_PATH }}
+      run: |
+        poetry run pytest --cov aiida_mlip --cov-append .
```
ElliottKasoar, line 58:
```suggestion
      run: poetry run pytest --cov aiida_mlip --cov-append .
```
Very minor (and I think introduced by me) but looks a bit messy

aiida_mlip/calculations/singlepoint.py
```diff
@@ -0,0 +1,271 @@
...
+            help="mlip model used for calculation",
+        )
+        spec.input("structure", valid_type=StructureData, help="The input structure.")
+        spec.input("precision", valid_type=Str, help="Precision level for calculation")
+        spec.input(
+            "device",
+            valid_type=Str,
+            required=False,
+            default=lambda: Str("cpu"),
+            help="Device in which to run calculation(cpu, gpu...)",
```
ElliottKasoar, line 80:
```suggestion
            help="Device on which to run calculation (cpu, cuda or mps)",
```
Probably misleading to say "gpu" anyway

tests/conftest.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+@pytest.fixture(scope="session")
+def filepath_tests():
+    """Return the absolute filepath of the `tests` folder.
+
+    .. warning:: if this file moves with respect to the `tests` folder, the
+    implementation should change.
+
+    :return: absolute filepath of `tests` folder which is the basepath for all test
```
ElliottKasoar, line 31: Are these supposed to be differently styled?

aiida_mlip/calculations/singlepoint.py
```diff
@@ -0,0 +1,271 @@
...
+        cif_structure = structure.get_cif()
+        with folder.open(self._DEFAULT_INPUT_FILE, "w", encoding="utf-8") as inputfile:
+            inputfile.write(cif_structure.get_content())
+
+        cmd_line = {
+            "arch": architecture,
+            "struct": input_filename,
+            "device": device,
+            "log": log_filename,
+            "calc-kwargs": {"model_paths": model_path, "default_dtype": precision},
```
ElliottKasoar, line 241:
```suggestion
            "calc-kwargs": {"model": model_path, "default_dtype": precision},
```
Hopefully works either way

tests/calculations/test_singlepoint.py
```diff
@@ -0,0 +1,117 @@
...
+        "--arch",
+        "mace_mp",
+        "--struct",
+        "aiida.cif",
+        "--device",
+        "cpu",
+        "--log",
+        "aiida.log",
+        "--calc-kwargs",
+        f"{{'model_paths': '{tmp_path}/mace_mp/mace_mp_small.model', 'default_dtype': 'float64'}}",
```
ElliottKasoar, line 46:
```suggestion
        f"{{'model': '{tmp_path}/mace_mp/mace_mp_small.model', 'default_dtype': 'float64'}}",
```

aiida_mlip/parsers.py
```diff
@@ -1,58 +1,136 @@
...
 
-class DiffParser(Parser):
+def convert_numpy(dictionary: dict) -> dict:
+    """
+    A function to convert numpy ndarrays in dictionary into lists.
+
+    Parameters
+    ----------
+    dictionary : dict
+        A dictionary.
```
ElliottKasoar, line 26:
```suggestion
        A dictionary with numpy array values to be converted into lists.
```

aiida_mlip/parsers.py
```diff
@@ -1,58 +1,136 @@
...
+
+    Returns
+    -------
+    dict
+        A dictionary.
```
ElliottKasoar, line 31:
```suggestion
        Converted dictionary.
```

README.md
```diff
@@ -18,60 +18,36 @@ intended to help developers get started with their AiiDA plugins.
 * [`.github/`](.github/): [Github Actions](https://github.com/features/actions) configuration
   * [`ci.yml`](.github/workflows/ci.yml): runs tests, checks test coverage and builds documentation at every new commit
   * [`publish-on-pypi.yml`](.github/workflows/publish-on-pypi.yml): automatically deploy git tags to PyPI - just generate a [PyPI API token](https://pypi.org/help/#apitoken) for your PyPI account and add it to the `pypi_token` secret of your github repository
+  * [`docs.yml`](.github/workflows/docs.yml): builds and deploys the documentasion
```
ElliottKasoar, line 21:
```suggestion
  * [`docs.yml`](.github/workflows/docs.yml): builds and deploys the documentation
```

README.md
```diff
@@ -18,60 +18,36 @@ intended to help developers get started with their AiiDA plugins.
...
+  * [`calculations/`](aiida_mlip/calculations/): Plugin `Calcjob` classes
+    * [`singlepoint.py](aiida_mlip/calculations/singlepoint.py ): `Calcjob` class to run single point calculations using mlips
+  * [`parsers.py`](aiida_mlip/parsers.py): `Parser` for the `Singlepoint` calculation
+* [`docs/`](docs/source/): Code documentation
+  * [`apidoc/`](docs/source/apidoc/): API documentation
+  * [`developer_guide/`](docs/source/developer_guide/): Documentation for developers
+  * [`user_guide`](docs/source/user_guide/): Documentation for users
+  * [`images`](docs/source/images/): Logos etc used in the documentation
+* [`examples/`](examples/): Examples for submitting calculations using this plugin
+  *[`calculations/submit_singlepoint.py`](examples/calculations/submit_singlepoint.py): Script for submitting a singlepoint calculation
```
ElliottKasoar, line 34:
```suggestion
  * [`calculations/submit_singlepoint.py`](examples/calculations/submit_singlepoint.py): Script for submitting a singlepoint calculation
```

README.md
```diff
@@ -18,60 +18,36 @@ intended to help developers get started with their AiiDA plugins.
...
 * [`tests/`](tests/): Basic regression tests using the [pytest](https://docs.pytest.org/en/latest/) framework (submitting a calculation, ...). Install `pip install -e .[testing]` and run `pytest`.
+  * [`conftest.py`](conftest.py): Configuration of fixtures for [pytest](https://docs.pytest.org/en/latest/)
+  * [`tests/calculations`](tests/calculations/test_singlepoint.py): Test `SinglePoint` calculation
+  * [`tests/data`](tests/data/test_model.py): Test `ModelData` type
```
ElliottKasoar, line 38:
```suggestion
  * [`calculations/`](tests/calculations): Test `SinglePoint` calculation
    * [`test_singlepoint.py`](tests/calculations/test_singlepoint.py): Calculations
  * [`data/`](tests/data): `ModelData`
    * [`test_model.py`](tests/data/test_model.py): Test `ModelData` type
```
May be worth previewing these

README.md
```diff
@@ -18,60 +18,36 @@ intended to help developers get started with their AiiDA plugins.
...
 * [`.gitignore`](.gitignore): Telling git which files to ignore
 * [`.pre-commit-config.yaml`](.pre-commit-config.yaml): Configuration of [pre-commit hooks](https://pre-commit.com/) that sanitize coding style and check for syntax errors. Enable via `pip install -e .[pre-commit] && pre-commit install`
-* [`.readthedocs.yml`](.readthedocs.yml): Configuration of documentation build for [Read the Docs](https://readthedocs.org/)
-* [`LICENSE`](LICENSE): License for your plugin
+* [`LICENSE`](LICENSE): License for the plugin
 * [`README.md`](README.md): This file
-* [`conftest.py`](conftest.py): Configuration of fixtures for [pytest](https://docs.pytest.org/en/latest/)
+* [`toc.ini`](tox.ini): File to set up tox
```
ElliottKasoar, line 43:
```suggestion
* [`tox.ini`](tox.ini): File to set up tox
```

README.md
```diff
@@ -18,60 +18,36 @@ intended to help developers get started with their AiiDA plugins.
...
 * [`pyproject.toml`](setup.json): Python package metadata for registration on [PyPI](https://pypi.org/) and the [AiiDA plugin registry](https://aiidateam.github.io/aiida-registry/) (including entry points)
```
ElliottKasoar, line 44:
```suggestion
* [`pyproject.toml`](pyproject.toml): Python package metadata for registration on [PyPI](https://pypi.org/) and the [AiiDA plugin registry](https://aiidateam.github.io/aiida-registry/) (including entry points)
```

examples/calculations/submit_singlepoint.py
```diff
@@ -0,0 +1,167 @@
...
+    if string is None:
+        model = None
+    elif (file_path := Path(string)).is_file():
+        model = ModelData.local_file(file_path, architecture=architecture)
+    else:
+        model = ModelData.download(string, architecture=architecture)
+    return model
+
+
+def load_structure(value: Union[str, Path, int, None]) -> StructureData:
```
ElliottKasoar, line 47: `value` -> `struct` or something more obvious?

examples/calculations/submit_singlepoint.py
```diff
@@ -0,0 +1,167 @@
...
+    """
+    Load a StructureData instance from the given input.
+
+    The input can be either a path to a structure file, a node PK (int),
+    or None. If the input is None, a default StructureData instance for NaCl
+    with a rocksalt structure will be created.
+
+    Parameters
+    ----------
+    value : str, int or None
```
ElliottKasoar, line 57:
```suggestion
    value : Union[str, Path, int, None]
```

examples/calculations/submit_singlepoint.py
```diff
@@ -0,0 +1,167 @@
...
+        structure = StructureData(ase=read(value))
+    else:
+        raise click.BadParameter(
+            f"Invalid input: {value}. Must be either node PK (int) or a valid \
+                path to a structure file."
+        )
+    return structure
+
+
+def singlepoint(params: dict):
```
ElliottKasoar, line 86: Type hint for return (tuple of specific types?)?

pyproject.toml
```diff
@@ -28,10 +28,13 @@ documentation = "https://stfc.github.io/aiida-mlip/"
 [tool.poetry.dependencies]
 python = "^3.9"
 aiida-core = "^2.5"
+ase = "^3.22.1"
 voluptuous = "^0.14"
+PyCifRW = "^4.4.6"
 
 [tool.poetry.group.dev.dependencies]
 coverage = {extras = ["toml"], version = "^7.4.1"}
+janus-core = "^0.1.0a1"
```
ElliottKasoar, line 37:
```suggestion
janus-core = "^0.1.0b2"
```

pyproject.toml
```diff
@@ -111,6 +112,7 @@ sections = ['FUTURE', 'STDLIB', 'THIRDPARTY', 'AIIDA', 'FIRSTPARTY', 'LOCALFOLDE
 known_aiida = ['aiida']
 
 [tool.numpydoc_validation]
+
```
ElliottKasoar, line 115:
```suggestion
```

ElliottKasoar (changes requested):

examples/calculations/submit_singlepoint.py
```diff
@@ -0,0 +1,167 @@
...
+@click.argument("codelabel", type=str)
+@click.option("--calctype", default="singlepoint", type=str)
+@click.option(
+    "--file",
+    default=None,
+    type=Union[str, Path, int],
+    help="Specify the structure (aiida node or path to a structure file)",
+)
+@click.option(
+    "--model", default=None, type=str, help="Specify path or url of the model to use"
```
ElliottKasoar, line 137:
```suggestion
    "--model", default=None, type=Union[str, Path], help="Specify path or url of the model to use"
```
Not sure how it handles a path if it can be either, but probably doesn't hurt?

examples/calculations/submit_singlepoint.py
alinelena: i think is fine for the single point to be only one structure... aiida has a different philosophy about these things. you can always combine it in a workflow.

examples/calculations/submit_singlepoint.py
```diff
@@ -0,0 +1,168 @@
...
+@click.option("--precision", default="float64", type=str)
+def cli(
+    codelabel, calctype, file, model, architecture, device, precision
+):  # pylint: disable=too-many-arguments
+    """Click interface."""
+    try:
+        code = Code.get_from_string(codelabel)
+    except NotExistent:
+        print(f"The code '{codelabel}' does not exist.")
+        sys.exit(1)
```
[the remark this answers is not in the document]
alinelena: this need to be fixed and deal with it properly... all one needs is a non zero code

examples/calculations/submit_singlepoint.py
alinelena: yes value is quite general and non informative

examples/calculations/submit_singlepoint.py
alinelena: yes needs refactor SinglePointCalculation I suggest

examples/calculations/submit_singlepoint.py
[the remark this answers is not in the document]
alinelena: personall I will raise SystemExit unless anyone has objections

aiida_mlip/calculations/singlepoint.py
```diff
@@ -0,0 +1,271 @@
...
+        )
+        spec.exit_code(
+            307,
+            "ERROR_CONTENT_OUTPUT_FILES",
+            message="The output file does not contain the right content",
+        )
+
+    @classmethod
+    def validate_inputs(
+        cls, value: dict, port_namespace: aiida.engine.processes.ports.PortNamespace
```
ElliottKasoar, line 149: `value`-> `inputs`? Especially in the context of dictionaries, value is misleading

aiida_mlip/calculations/singlepoint.py
```diff
@@ -0,0 +1,271 @@
...
+    ):
```
ElliottKasoar, line 150: Return type hint?

aiida_mlip/calculations/singlepoint.py
```diff
@@ -0,0 +1,271 @@
...
+
+        valid_calctypes = {"singlepoint", "geom opt"}
+        if "calctype" in value:
+            if str(value["calctype"].value) not in valid_calctypes:
+                return f"The 'calctype' must be one of {valid_calctypes}, \
+                    but got '{value['calctype']}'."
+
+        if "input_filename" in value:
+            if not str(value["input_filename"].value).endswith(".cif"):
+                return "The parameter 'input_filename' must end with '.cif'"
```
ElliottKasoar, line 184: Can these not be ValueErrors? It seems odd to return this as a string, but I'm guessing it might be how aiida expects it?

aiida_mlip/calculations/singlepoint.py
```diff
@@ -0,0 +1,271 @@
...
+        Default log file name.
+
+    Methods
+    -------
+    define(spec: CalcJobProcessSpec) -> None:
+        Define the process specification, its inputs, outputs and exit codes.
+    validate_inputs(value: dict, port_namespace: PortNamespace) -> Optional[str]:
+        Check if the inputs are valid.
+    prepare_for_submission(folder: Folder) -> CalcInfo:
+        Create the input files for the `Calcjob`.
```
ElliottKasoar, line 34:
```suggestion
        Create the input files for the `CalcJob`.
```

ElliottKasoar (changes requested):

aiida_mlip/calculations/singlepoint.py
alinelena: yap

aiida_mlip/calculations/singlepoint.py
```diff
@@ -0,0 +1,273 @@
...
+
+        spec.exit_code(
+            305,
+            "ERROR_MISSING_OUTPUT_FILES",
+            message="Some output files missing or cannot be read",
+        )
+        spec.exit_code(
+            306,
+            "ERROR_EMPTY_OUTPUT_FILES",
+            message="File.xyz is empty",
```
oerc0122, line 141: Is this file always called "File.xyz"? If not this is an unhelpful error message.

aiida_mlip/calculations/singlepoint.py
```diff
@@ -0,0 +1,273 @@
...
+        )
+        spec.exit_code(
+            307,
+            "ERROR_CONTENT_OUTPUT_FILES",
+            message="The output file does not contain the right content",
```
oerc0122, line 146: This doesn't sound like an issue users could fix. Certainly, I would have no idea how. Needs explaining

aiida_mlip/calculations/singlepoint.py
```diff
@@ -0,0 +1,273 @@
...
+        if any(key not in port_namespace for key in ("calctype", "structure")):
+            return None
+
+        for key in ("calctype", "structure"):
+            if key not in inputs:
+                return f"required value was not provided for the `{key}` namespace."
+
+        valid_calctypes = {"singlepoint", "geom opt"}
+        if "calctype" in inputs:
+            if str(inputs["calctype"].value) not in valid_calctypes:
```
oerc0122, line 180: Don't need to nest `if`s like this. Can combine with `and` instead which is short circuiting, so they are only executed while the previous one is true.
```suggestion
        if (
                "calctype" in inputs and
                str(inputs["calctype"].value) not in valid_calctypes
        ):
```

aiida_mlip/calculations/singlepoint.py
```diff
@@ -0,0 +1,273 @@
...
+                return f"The 'calctype' must be one of {valid_calctypes}, \
+                    but got '{inputs['calctype']}'."
```
oerc0122, line 182: Likewise, I believe the favoured way of concatenating long strings is to write:
```suggestion
                return (f"The 'calctype' must be one of {valid_calctypes}, "
                        f"but got '{inputs['calctype']}'.")
```

aiida_mlip/calculations/singlepoint.py
```diff
@@ -0,0 +1,273 @@
...
+            ).filepath
+        )
+
+        # The inputs are saved in the node, but we want their value as a string
+        calctype = str((self.inputs.calctype).value)
+        precision = str((self.inputs.precision).value)
+        device = str((self.inputs.device).value)
+        xyz_filename = str((self.inputs.xyz_output_name).value)
+        input_filename = self.inputs.metadata.options.input_filename
+        log_filename = str((self.inputs.log_filename).value)
```
oerc0122, line 231: Seems awkward to be constantly casting to strings. Would recommend adding getter `@property`s on the object for these. 

```python
@property
def calctype(self):
    return str(self.inputs.calctype.value)
```

aiida_mlip/data/model.py
```diff
@@ -9,15 +9,57 @@
 from aiida.orm import SinglefileData
 
 
+# pylint : disable = line-too-long
```
oerc0122, line 12: Does this need to be disabled globally?

aiida_mlip/parsers.py
```diff
@@ -1,58 +1,136 @@
...
         # Note: set(A) <= set(B) checks whether A is a subset of B
         if not set(files_expected) <= set(files_retrieved):
             self.logger.error(
                 f"Found files '{files_retrieved}', expected to find '{files_expected}'"
             )
             return self.exit_codes.ERROR_MISSING_OUTPUT_FILES
 
-        # add output file
-        self.logger.info(f"Parsing '{output_filename}'")
+        # Add output file to the outpus
```
oerc0122, line 112:
```suggestion
        # Add output file to the outputs
```

aiida_mlip/parsers.py
```diff
@@ -1,58 +1,136 @@
...
+        xyzoutput = (self.node.inputs.xyz_output_name).value
+        logoutput = (self.node.inputs.log_filename).value
+
+        remote_folder = self.node.get_remote_workdir()
 
         # Check that folder content is as expected
         files_retrieved = self.retrieved.list_object_names()
-        files_expected = [output_filename]
+
+        files_expected = [xyzoutput, logoutput]
```
oerc0122, line 104: Instead of casting to `set` you can just construct a `set`
```suggestion
        files_expected = {xyzoutput, logoutput}
```

aiida_mlip/parsers.py
```diff
@@ -1,58 +1,136 @@
...
+        node : aiida.orm.nodes.process.process.ProcessNode
+            ProcessNode of calculation.
         """
         super().__init__(node)
-        if not issubclass(node.process_class, DiffCalculation):
-            raise exceptions.ParsingError("Can only parse DiffCalculation")
+        if not issubclass(node.process_class, Singlepointcalc):
+            raise exceptions.ParsingError("Can only parse Singlepointcalc")
 
+    # pylint: disable=too-many-locals
```
oerc0122, line 80: Not sure we want this, much as I hate to say it, I kind of agree with pylint here. You have a lot of single use variables which don't really add much clarity to me.

aiida_mlip/parsers.py
```diff
@@ -1,58 +1,136 @@
...
+            xyz_node = SinglefileData(file=handle)
+
+        self.out("xyz_output", xyz_node)
+
         with self.retrieved.open(output_filename, "rb") as handle:
-            output_node = SinglefileData(file=handle)
-        self.out("mlip", output_node)
+            stdout_node = SinglefileData(file=handle)
+
+        self.out("std_output", stdout_node)
```
oerc0122, line 128: Would this work?
```suggestion
        with self.retrieved.open(logoutput, "rb") as handle:
            self.out("log_output", SinglefileData(file=handle))

        with self.retrieved.open(xyzoutput, "rb") as handle:
            self.out("xyz_output", SinglefileData(file=handle))

        with self.retrieved.open(output_filename, "rb") as handle:
            self.out("std_output",  SinglefileData(file=handle))
```
With 3.9+ (parens on with) you could do this as:
```suggestion
        with (self.retrieved.open(logoutput, "rb") as log_handle,
              self.retrieved.open(xyzoutput, "rb") as xyz_handle,
              self.retrieved.open(output_filename, "rb") as out_handle):
            self.out("log_output", SinglefileData(file=log_handle))
            self.out("xyz_output", SinglefileData(file=xyz_handle))
            self.out("std_output",  SinglefileData(file=out_handle))
```

aiida_mlip/parsers.py
oerc0122: This is also single use and could be a constant in the `if` below

README.md
```diff
@@ -90,16 +68,10 @@ A quick demo of how to submit a calculation:
 ```shell
 verdi daemon start     # make sure the daemon is running
 cd examples
-./example_01.py        # run test calculation
+verdi run submit_singlepoint.py "janus@localhost" --calctype "singlepoint"  --architecture mace_mp --model "~./cache/mlips/mace_mp/46jrkm3v"       # run test calculation
```
alinelena, line 71: this shall work with mace... makes it more general anyhow

README.md
```diff
@@ -18,60 +18,38 @@ intended to help developers get started with their AiiDA plugins.
...
- * [aiida-diff tests](https://www.youtube.com/watch?v=2CxiuiA1uVs&t=1931s)
- * [Adding your plugin to the registry](https://www.youtube.com/watch?v=760O2lDB-TM&t=112s)
- * [pre-commit hooks](https://www.youtube.com/watch?v=760O2lDB-TM&t=333s)
-
-For more information, see the [developer guide](https://aiida-diff.readthedocs.io/en/latest/developer_guide) of your plugin.
 
 
 ## Features
 
- * Add input files using `SinglefileData`:
```
alinelena, line 55: see the list of features in janus-core and add a similar bullet one here, for the moment only single point, geometry optimisation, and md, training and fine tunning probably too.

docs/source/conf.py
```diff
@@ -44,12 +44,16 @@
     "sphinxcontrib.contentui",
 ]
 
+# Add the numpydoc_show_inherited_class_members option
+numpydoc_show_inherited_class_members = False
```
oerc0122, line 48: This might fix the issues we had with `TypedDict` too

examples/calculations/submit_singlepoint.py
```diff
@@ -0,0 +1,167 @@
...
+    if string is None:
+        model = None
+    elif (file_path := Path(string)).is_file():
+        model = ModelData.local_file(file_path, architecture=architecture)
+    else:
+        model = ModelData.download(string, architecture=architecture)
+    return model
+
+
+def load_structure(struct: Union[str, Path, int, None]) -> StructureData:
```
oerc0122, line 46: This might be a useful place for [`functools.singledispatch`](https://docs.python.org/3/library/functools.html#functools.singledispatch), but then again, I always think that.

tests/conftest.py
```diff
@@ -0,0 +1,114 @@
...
+    """Return the absolute filepath to the directory containing the file `fixtures`."""
+    return Path(filepath_tests / "fixtures")
+
+
+@pytest.fixture(scope="function")
+def fixture_sandbox():
+    """Return a `SandboxFolder`."""
+
+    with SandboxFolder() as folder:
+        yield folder
```
oerc0122, line 48: Any reason this `yield`s rather than `return`ing?

tox.ini
```diff
@@ -6,9 +6,9 @@ usedevelop=True
 
 [testenv:py{39,310,311,312}]
 description = Run the test suite against Python versions
-allowlist_externals = poetry
+allowlist_externals = poetry, sh
 commands_pre = poetry install --no-root --sync
-commands = poetry run pytest --cov aiida_mlip --import-mode importlib
+commands = sh -c 'export JANUS_PATH=$(which janus); exec "$@"' _ poetry run pytest --cov aiida_mlip --import-mode importlib
```
oerc0122, line 11: A comment explaining the shell magic here might be nice.

oerc0122 (changes requested):

tests/conftest.py
federicazanca: no I copied it from other and forgot to change , I'll change it

docs/source/conf.py
alinelena: it may be... @ElliottKasoar

tox.ini
alinelena: @ElliottKasoar can you help in here

examples/calculations/submit_singlepoint.py
alinelena: this is an example let us keep it simple for the normal user...

README.md
```diff
@@ -18,60 +18,38 @@ intended to help developers get started with their AiiDA plugins.
...
+  * [`parsers.py`](aiida_mlip/parsers.py): `Parser` for the `Singlepoint` calculation
+* [`docs/`](docs/source/): Code documentation
+  * [`apidoc/`](docs/source/apidoc/): API documentation
+  * [`developer_guide/`](docs/source/developer_guide/): Documentation for developers
+  * [`user_guide/`](docs/source/user_guide/): Documentation for users
+  * [`images/`](docs/source/images/): Logos etc used in the documentation
+* [`examples/`](examples/): Examples for submitting calculations using this plugin
+  * [`calculations/submit_singlepoint.py`](examples/calculations/submit_singlepoint.py): Script for submitting a singlepoint calculation
 * [`tests/`](tests/): Basic regression tests using the [pytest](https://docs.pytest.org/en/latest/) framework (submitting a calculation, ...). Install `pip install -e .[testing]` and run `pytest`.
+  * [`conftest.py`](conftest.py): Configuration of fixtures for [pytest](https://docs.pytest.org/en/latest/)
```
ElliottKasoar, line 36:
```suggestion
  * [`conftest.py`](tests/conftest.py): Configuration of fixtures for [pytest](https://docs.pytest.org/en/latest/)
```

aiida_mlip/parsers.py
```diff
@@ -1,58 +1,136 @@
...
+            ProcessNode of calculation.
         """
         super().__init__(node)
-        if not issubclass(node.process_class, DiffCalculation):
-            raise exceptions.ParsingError("Can only parse DiffCalculation")
+        if not issubclass(node.process_class, Singlepointcalc):
+            raise exceptions.ParsingError("Can only parse Singlepointcalc")
 
+    # pylint: disable=too-many-locals
     def parse(self, **kwargs):
```
ElliottKasoar, line 81:
```suggestion
    def parse(self, **kwargs) -> int:
```

aiida_mlip/parsers.py
```diff
@@ -1,58 +1,136 @@
...
         """
-        Parse outputs, store results in database.
+        Parse outputs, store results in the database.
+
+        Parameters
+        ----------
+        **kwargs : Any
+            Any keyword argument.
```
ElliottKasoar, line 88:
```suggestion
            Any keyword arguments.
```

tests/calculations/test_singlepoint.py
```diff
@@ -0,0 +1,117 @@
...
+            cache_dir=tmp_path,
+        ),
+        "device": Str("cpu"),
+    }
+    with pytest.raises(ValueError):
+        generate_calc_job(fixture_sandbox, entry_point_name, inputs)
+
+
+def test_run_sp(tmp_path, janus_code):
+    """Test singlepoint calculation"""
```
ElliottKasoar, line 91:
```suggestion
def test_singlepoint(fixture_sandbox, generate_calc_job, tmp_path, janus_code):
    """Test generating singlepoint calculation job"""
    # pylint:disable=line-too-long
    entry_point_name = "janus.sp"
    inputs = {
        "metadata": {"options": {"resources": {"num_machines": 1}}},
        "code": janus_code,
        "architecture": Str("mace_mp"),
        "precision": Str("float64"),
        "structure": StructureData(ase=bulk("NaCl", "rocksalt", 5.63)),
        "calctype": Str("singlepoint"),
        "model": ModelData.download(
            "https://github.com/stfc/janus-core/raw/main/tests/models/mace_mp_small.model",
            architecture="mace_mp",
            cache_dir=tmp_path,
        ),
        "device": Str("cpu"),
    }

    calc_info = generate_calc_job(fixture_sandbox, entry_point_name, inputs)
    # pylint:disable=line-too-long
    cmdline_params = [
        "singlepoint",
        "--arch",
        "mace_mp",
        "--struct",
        "aiida.cif",
        "--device",
        "cpu",
        "--log",
        "aiida.log",
        "--calc-kwargs",
        f"{{'model': '{tmp_path}/mace_mp/mace_mp_small.model', 'default_dtype': 'float64'}}",
        "--write-kwargs",
        "{'filename': 'aiida-results.xyz'}",
    ]

    retrieve_list = [
        calc_info.uuid,
        "aiida.log",
        "aiida-results.xyz",
        "aiida-stdout.txt",
    ]

    print(sorted(cmdline_params))
    print(sorted(calc_info.codes_info[0].cmdline_params))
    # Check the attributes of the returned `CalcInfo`
    assert isinstance(calc_info, datastructures.CalcInfo)
    assert isinstance(calc_info.codes_info[0], datastructures.CodeInfo)
    assert sorted(calc_info.codes_info[0].cmdline_params) == sorted(cmdline_params)
    assert sorted(calc_info.retrieve_list) == sorted(retrieve_list)
    assert sorted(fixture_sandbox.get_content_list()) == ["aiida.cif"]


def test_sp_error(fixture_sandbox, generate_calc_job, tmp_path, janus_code):
    """Test singlepoint calculation with error input"""
    entry_point_name = "janus.sp"
    # pylint:disable=line-too-long
    inputs = {
        "metadata": {"options": {"resources": {"num_machines": 1}}},
        "code": janus_code,
        "architecture": Str("mace_mp"),
        "precision": Str("float64"),
        "structure": StructureData(ase=bulk("NaCl", "rocksalt", 5.63)),
        "calctype": Str("wrong_type"),
        "model": ModelData.download(
            "https://github.com/stfc/janus-core/raw/main/tests/models/mace_mp_small.model",
            architecture="mace_mp",
            cache_dir=tmp_path,
        ),
        "device": Str("cpu"),
    }
    with pytest.raises(ValueError):
        generate_calc_job(fixture_sandbox, entry_point_name, inputs)


def test_run_sp(tmp_path, janus_code):
    """Test running singlepoint calculation"""
```
(There may be a more accurate description, but aiming to differentiate these tests)

ElliottKasoar (changes requested):

aiida_mlip/calculations/singlepoint.py
federicazanca: good point and I don't even need the str actually (value is already a str) , so I'll remove it

tox.ini
ElliottKasoar, line 11:
```suggestion
# Export path to janus executable installed in tox environment before running pytest
commands = sh -c 'export JANUS_PATH=$(which janus); exec "$@"' _ poetry run pytest --cov aiida_mlip --import-mode importlib
```

aiida_mlip/calculations/singlepoint.py
```diff
@@ -0,0 +1,273 @@
...
+            else str(self.inputs.architecture.value)
+        )
+        model_path = (
+            str((self.inputs.model).filepath)
+            if self.inputs.model
+            else ModelData.download(
+                "https://github.com/stfc/janus-core/raw/main/tests/models/mace_mp_small.model",  # pylint:disable=line-too-long
+                architecture,
+            ).filepath
+        )
```
ElliottKasoar, line 217:
```suggestion
        if self.inputs.model:
            model_path = self.inputs.model.filepath
        else:
            model_path = ModelData.download(
                "https://github.com/stfc/janus-core/raw/main/tests/models/mace_mp_small.model",  # pylint:disable=line-too-long
                architecture,
            ).filepath
```
Somewhat down to personal preference, but I generally avoid "one-liner" if/else if it spreads over multiple lines.

**Note**: I've also removed `str` conversions, as I don't think they're needed (tests pass for me) but that may be wrong

aiida_mlip/calculations/singlepoint.py
```diff
@@ -0,0 +1,267 @@
...
+                )
+
+        valid_calctypes = {"singlepoint", "geom opt"}
+        if (
+            "calctype" in inputs
+            and str(inputs["calctype"].value) not in valid_calctypes
+        ):
+            raise ValueError(
+                f"The 'calctype' must be one of {valid_calctypes}, \
+                    but got '{inputs['calctype']}'."
```
ElliottKasoar, line 178:
```suggestion
                f"The 'calctype' must be one of {valid_calctypes}, "
                f"but got '{inputs['calctype']}'."
```

aiida_mlip/data/model.py
```diff
@@ -10,14 +10,55 @@
...
+    set_file(file: Union[str, Path], filename: Optional[str] = None, 
+        architecture: Optional[str] = None, **kwargs: Any) -> None:
+        Set the file for the node.
+    local_file(file: Union[str, Path], architecture: str, 
+        filename: Optional[str] = None) -> 'ModelData':
+        Create a ModelData instance from a local file.
+    download(url: str, architecture: str, filename: Optional[str] = None, 
+        cache_dir: Optional[Union[str, Path]] = None, 
+            force_download: Optional[bool] = False) -> 'ModelData':
+        Download a file from a URL and save it as ModelData.
```
ElliottKasoar, line 50:
```suggestion
    _calculate_hash(file: Union[str, Path]) -> str:
        Calculate the hash of a file.
    _check_existing_file(file: Union[str, Path]) -> Path:
        Check if a file already exists and return the path of the existing file.
    __init__(file: Union[str, Path], architecture: str,
        filename: Optional[str] = None, **kwargs: Any) -> None:
        Initialize the ModelData object.
    set_file(file: Union[str, Path], filename: Optional[str] = None,
        architecture: Optional[str] = None, **kwargs: Any) -> None:
        Set the file for the node.
    local_file(file: Union[str, Path], architecture: str,
        filename: Optional[str] = None) -> 'ModelData':
        Create a ModelData instance from a local file.
    download(url: str, architecture: str, filename: Optional[str] = None,
        cache_dir: Optional[Union[str, Path]] = None,
            force_download: Optional[bool] = False) -> 'ModelData':
        Download a file from a URL and save it as ModelData.
```

aiida_mlip/data/model.py
```diff
@@ -36,17 +35,17 @@ class ModelData(SinglefileData):
...
-    local_file(file: Union[str, Path], architecture: str, \
+    local_file(file: Union[str, Path], architecture: str, 
         filename: Optional[str] = None) -> 'ModelData':
         Create a ModelData instance from a local file.
-    download(url: str, architecture: str, filename: Optional[str] = None, \
-        cache_dir: Optional[Union[str, Path]] = None, \
+    download(url: str, architecture: str, filename: Optional[str] = None, 
+        cache_dir: Optional[Union[str, Path]] = None, 
             force_download: Optional[bool] = False) -> 'ModelData':
         Download a file from a URL and save it as ModelData.
```
ElliottKasoar, line 50: It won't let me suggest this, but I think you can replace all of this with

```
    Methods
    -------
    set_file(file, filename=None, architecture=None, **kwargs)
        Set the file for the node.
    local_file(file, architecture, filename=None):
        Create a ModelData instance from a local file.
    download(url, architecture, filename=None, cache_dir=None, force_download=False)
        Download a file from a URL and save it as ModelData.
```
In general you don't need to do the type hints/non-public functions for this section: https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring

README.md
```diff
@@ -18,67 +18,60 @@ intended to help developers get started with their AiiDA plugins.
...
- * [aiida-diff computer/code helpers](https://www.youtube.com/watch?v=2CxiuiA1uVs&t=1238s)
- * [aiida-diff input data (with validation)](https://www.youtube.com/watch?v=2CxiuiA1uVs&t=1353s)
- * [aiida-diff cli](https://www.youtube.com/watch?v=2CxiuiA1uVs&t=1621s)
- * [aiida-diff tests](https://www.youtube.com/watch?v=2CxiuiA1uVs&t=1931s)
- * [Adding your plugin to the registry](https://www.youtube.com/watch?v=760O2lDB-TM&t=112s)
- * [pre-commit hooks](https://www.youtube.com/watch?v=760O2lDB-TM&t=333s)
 
-For more information, see the [developer guide](https://aiida-diff.readthedocs.io/en/latest/developer_guide) of your plugin.
+## Features
+## Features (this is in development not all things work yet)
```
ElliottKasoar, line 51:
```suggestion
## Features (in development)
```

tests/calculations/test_singlepoint.py
```diff
@@ -0,0 +1,138 @@
...
+        "metadata": {"options": {"resources": {"num_machines": 1}}},
+        "code": janus_code,
+        "architecture": Str("mace_mp"),
+        "precision": Str("float64"),
+        "structure": StructureData(ase=bulk("NaCl", "rocksalt", 5.63)),
+        "calctype": Str("singlepoint"),
+        "model": ModelData.download(
+            "https://github.com/stfc/janus-core/raw/main/tests/models/mace_mp_small.model",
+            architecture="mace_mp",
+            cache_dir=tmp_path,
```
ElliottKasoar, line 30: I haven't looked into it yet, but if you replace `tmp_path` with `str(tmp_path)` this fails (`TypeError: expected str, bytes or os.PathLike object, not NoneType`) which suggests that the str version doesn't work.

aiida_mlip/data/model.py
```diff
@@ -30,24 +29,13 @@ class ModelData(SinglefileData):
     filepath : str
         Path of the mlip model.
 
-    Methods
+     Methods
```
oerc0122, line 32: Is this meant to be indented by one?

tests/calculations/test_singlepoint.py
ElliottKasoar: Doesn't seem to be a problem actually, works either way now.

tests/conftest.py
```diff
@@ -23,55 +23,117 @@ def clear_database_auto(aiida_profile_clean):  # pylint: disable=unused-argument
...
+    Returns
+    -------
+    Path
+        Absolute filepath of `tests` folder.
+
+    Warns
+    -----
+    Warning
+        If this file moves with respect to the `tests` folder,
+        the implementation should change.
```
oerc0122, line 42: This does not issue a warning. This should be a 
```
.. warning:
```
in the summary

tests/conftest.py
```diff
@@ -112,3 +196,18 @@ def _generate_calc_job(folder, entry_point_name, inputs=None):
...
+    """
+    Fixture to provide the path to the example file.
+
+    Returns:
+        Path: The path to the example file.
+    """
+    return Path(__file__).resolve().parent.parent
+           / "examples"
+           / "calculations"
+           / "submit_singlepoint.py"
```
oerc0122: Does this work? The new lines make me think it might not.

alinelena: nope this is old fixed already... actually wants a tuple hence the yield above....

tests/calculations/test_singlepoint.py
```diff
@@ -37,15 +37,15 @@ def test_singlepoint(fixture_sandbox, generate_calc_job, tmp_path, janus_code):
...
+        "mace",
         "--struct",
         "aiida.cif",
         "--device",
         "cpu",
         "--log",
         "aiida.log",
         "--calc-kwargs",
-        f"{{'model': '{tmp_path}/mace_mp/mace_mp_small.model', 'default_dtype': 'float64'}}",
+        f"{{'model': '{tmp_path}/mace/mace_mp_small.model', 'default_dtype': 'float64'}}",
```
oerc0122, line 48:
```suggestion
        f"{{'model': '{tmp_path}/mace/mace_small.model', 'default_dtype': 'float64'}}",
```
?

oerc0122 (approved):

alinelena (approved):

oerc0122 (dismissed):

ElliottKasoar (approved):
There's a mix of `singlePointCalculation` and `Singlepointcalc` between test_singlepoint.py and parsers.py, neither of which I'd necessarily choose (maybe `SinglePointCalc`), but I don't think it's a problem.

oerc0122 (approved):

ElliottKasoar (approved):

Issue #86, opened 2024-03-01
Harmonize max line length


ElliottKasoar:
See https://github.com/stfc/janus-core/pull/50

Currently isort and black have different maximum line lengths, so can clash in some cases, preventing the pre-commit passing. As long as they agree, it should be fine, but I'm happy with black's default of 88.

ElliottKasoar:
Possibly worth doing this at the same time as #85, since if we set a pylint max line length of 88, some docstrings need changing anyway.

PR #90, opened 2024-03-08
Update CI

### LINKED ISSUE #85: Add numpydoc Sphinx extension
See https://github.com/stfc/janus-core/pull/44 for possible implementation

ElliottKasoar:
Resolves #85 and #86

Based on equivalent changes in janus-core (see linked issues).

Currently fails due to existing code base not conforming to these standards.

It might be worth integrating this into #84 @federicazanca, rather than me rewriting the existing docstrings etc. that may change now, and then rewriting more once the new code is merged?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,7 +79,7 @@ jobs:
     - name: Set up Python 3.11
       uses: actions/setup-python@v5
       with:
-        python-version: 3.11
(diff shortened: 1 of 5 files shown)


alinelena (approved):

alinelena:
this is an exception and will cause ci to fail... in normal circumstances we shall not merge in such a state

PR #98, opened 2024-03-19
Fix pytest and tox

ElliottKasoar:
- Fixes potential errors if utility modules are added to `tests`
- Fixes pytest error if `JANUS_PATH` is not set but janus is in the PATH
- Allows arguments to be passed to pytest via tox e.g. specifying specific tests to run

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -9,7 +9,7 @@ description = Run the test suite against Python versions
 allowlist_externals = poetry, sh
 commands_pre = poetry install --no-root --sync
 # Export path to janus executable installed in tox environment before running pytest
-commands = sh -c 'export JANUS_PATH=$(which janus); exec "$@"' _ poetry run pytest --cov aiida_mlip --import-mode importlib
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -9,7 +9,7 @@ description = Run the test suite against Python versions
 allowlist_externals = poetry, sh
 commands_pre = poetry install --no-root --sync
 # Export path to janus executable installed in tox environment before running pytest
-commands = sh -c 'export JANUS_PATH=$(which janus); exec "$@"' _ poetry run pytest --cov aiida_mlip --import-mode importlib
+commands = sh -c 'export JANUS_PATH=$(which janus); exec "$@"' _ poetry run pytest {posargs} --cov aiida_mlip --import-mode importlib
```
alinelena, line 12: now with your fix, probably the export will not be needed anymore, can you double check?

ElliottKasoar: Seems to, I've removed it from the CI too which hopefully also works

alinelena (approved):

PR #107, opened 2024-04-10
Logo

alinelena:


ElliottKasoar:
Should we not have the logo in the [docs/source/index](https://github.com/stfc/aiida-mlip/blob/main/docs/source/index.rst) too?

(I know it's not technically in the scope, but might be nice to fix the BSD 3-Clause licence link while we're touching the file.)

ElliottKasoar (approved):

ElliottKasoar (approved):

PR #116, opened 2024-04-29
fix singlefiledata and some names

federicazanca:
just some small changes to make provenance and some names look better

### DIFF
--- a/examples/calculations/submit_geomopt.py
+++ b/examples/calculations/submit_geomopt.py
@@ -27,7 +27,7 @@ def geomopt(params: dict) -> None:
     structure = load_structure(params["struct"])
 
     # Select model to use
-    model = load_model(params["model"], params["arch"])
+    model = load_model(params["model"], params["architecture"])
 
     # Select calculation to use
     geomoptCalculation = CalculationFactory("janus.opt")
@@ -36,18 +36,17 @@ def geomopt(params: dict) -> None:
     inputs = {
         "metadata": {"options": {"resources": {"num_machines": 1}}},
         "code": params["code"],
-        "arch": Str(params["arch"]),
-        "struct": structure,
+        "architecture": Str(params["architecture"]),
+        "structure": structure,
+        "model": model,
         "precision": Str(params["precision"]),
         "device": Str(params["device"]),
-        "fmax": Float(params["fmax"]),
+        "max_force": Float(params["max_force"]),
         "vectors_only": Bool(params["vectors_only"]),
         "fully_opt": Bool(params["fully_opt"]),
         "opt_kwargs": Dict({"restart": "rest.pkl"}),
         "steps": Int(params["steps"]),
     }
-    if model is not None:
-        inputs["model"] = model
 
     # Run calculation
     result, node = run_get_node(geomoptCalculation, **inputs)
@@ -71,7 +70,7 @@ def geomopt(params: dict) -> None:
     help="Specify path or url of the model to use",
 )
 @click.option(
-    "--arch",
+    "--architecture",
     default="mace_mp",
     type=str,
     help="MLIP architecture to use for calculations.",
@@ -82,7 +81,9 @@ def geomopt(params: dict) -> None:
 @click.option(
     "--precision", default="float64", type=str, help="Chosen level of precision."
 )
-@click.option("--fmax", default=0.1, type=float, help="Maximum force for convergence.")
+@click.option(
+    "--max_force", default=0.1, type=float, help="Maximum force for convergence."
+)
 @click.option(
     "--vectors_only",
     default=False,
@@ -102,10 +103,10 @@ def cli(
     codelabel,
     struct,
     model,
-    arch,
+    architecture,
     device,
     precision,
-    fmax,
+    max_force,
     vectors_only,
     fully_opt,
     steps,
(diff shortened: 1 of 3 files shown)


examples/calculations/submit_singlepoint.py
```diff
@@ -36,12 +36,13 @@ def singlepoint(params: dict) -> None:
...
         "metadata": {"options": {"resources": {"num_machines": 1}}},
         "code": params["code"],
-        "architecture": Str(params["architecture"]),
+        "arch": Str(params["arch"]),
         "structure": structure,
-        "model": model,
         "precision": Str(params["precision"]),
         "device": Str(params["device"]),
     }
+    if model is not None:
```
ElliottKasoar, line 44: Do we not need a model by this point? (If there's a default, would it not be defined in `load_model`?)

(This applies in the other examples too)

federicazanca: there is a default in the calcjob code itself, so the load_model function I made it to return None if None was given as an argument.
But if I input model=None, aiida explodes so I did it this way. 
Probably it owuld be better to have a default in the load_model, the problem is that the load_model wants a url or a path, and if I default a path, then another person using it might not have that path.
Idk if it makes sense

federicazanca:
> > just some small changes to make provenance and some names look better
> 
> Just to clarify, look better in what sense?

Just when I was making the provenance graph for my calculations one of the reasons it looked very large and unreadable is that the singlefiledata types default the "filename" to a very long string, while now it's a name that makes more sense.
see pictures (the outputs in particular)for comparison. 
image
image

federicazanca:
This PR should be only for changing the singlefiledata filename so if you are ok with those we can approve. 
I did some small changes in the model part that elliot was commenting but I would say if we really don't like how that's handled I open another issue about that?

examples/calculations/submit_singlepoint.py
ElliottKasoar: Whether we have a default or not is probably a separate question (and I don't have a strong opinion).

My question for now is, since we only set `inputs["model"]` if `model is not None`, what does happen if `model is None` here? In particular, does it give a useful error/warning or still explode in some form?

examples/calculations/submit_geomopt.py
```diff
@@ -27,7 +27,7 @@ def geomopt(params: dict) -> None:
     structure = load_structure(params["struct"])
 
     # Select model to use
-    model = load_model(params["model"], params["architecture"])
+    model = load_model(params["model"], params["arch"])
```
alinelena, line 30: this needs to be fixed separately not in this.

examples/calculations/submit_singlepoint.py
federicazanca: if you set it as None the calcjob accept it as an input and when later it has to do some stuff with it it explode cause it will look for some properties of the ModelData that `None` does not have

ElliottKasoar: Sorry, I may be misunderstanding, but with your current change, the model is no longer in the `inputs`, so would something different not happen? Or is `model` itself still passed even if it's not in the `inputs` dictionary?

I don't think it makes sense to let it get far enough to explode, but even more than before, I think this should be addressed separately.

aiida_mlip/parsers/base_parser.py
```diff
@@ -80,7 +80,9 @@ def parse(self, **kwargs) -> int:
             self.retrieved.open(log_output, "rb") as log,
             self.retrieved.open(output_filename, "rb") as output,
         ):
-            self.out("log_output", SinglefileData(file=log))
-            self.out("std_output", SinglefileData(file=output))
+            self.out("log_output", SinglefileData(file=log, filename=log_output))
```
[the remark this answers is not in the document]
federicazanca: I'm not sure I understood what you suggest me to do, but if it is something to change in the singlefiledata, that's part of aiida and I can't

examples/calculations/submit_singlepoint.py
federicazanca: If I don't give a model, it will use a default model that is defined inside the calcjob. If I give None, it will take None as model and explode. I have no idea if this makes sense (I would think None and giving nothing would act the same) but that's what it does

alinelena:
is fine if explodes but shall explode with an error.

examples/calculations/submit_geomopt.py
federicazanca: yea I know just I changed them cause I thought something was exploding because of these after modifying the singledata (obviously they were actually not the cause) and decided to leave the changes

federicazanca:
> is fine if explodes but shall explode with an error.

referring to the model thing? 
sure, I suggest we close this and I open another pr for making the model handling a bit better (and in general I'll try to go through some similar bugs that maybe I didn't notice initially

federicazanca:
> > > > just some small changes to make provenance and some names look better
> > > 
> > > 
> > > Just to clarify, look better in what sense?
> > 
> > 
> > Just when I was making the provenance graph for my calculations one of the reasons it looked very large and unreadable is that the singlefiledata types default the "filename" to a very long string, while now it's a name that makes more sense. see pictures (the outputs in particular)for comparison.
> 
> Ok thanks. Are we still looking at consolidating the nodes at some point, separately?

yea I think

ElliottKasoar:
I've reverted the changes so this is now just the changes for the SinglefileData filenames. We should still decide about wheher to use [.name](https://docs.python.org/3/library/io.html#io.FileIO.name).

I'll add the other changes in separate PRs.

alinelena (approved):

PR #123, opened 2024-05-15
121 add training

### LINKED ISSUE #121: Add Training

federicazanca:
just a messy draft, still need to 
- [x] check if it works
- [x] add tests
- [x] docs
closes #121


aiida_mlip/calculations/train.py
```diff
@@ -0,0 +1,170 @@
...
+    ----------
+    inputs : dict
+        The inputs dictionary.
+
+    port_namespace : `aiida.engine.processes.ports.PortNamespace`
+        An instance of aiida's `PortNameSpace`.
+
+    Raises
+    ------
+    ValueError
```
ElliottKasoar, line 31:
```suggestion
    InputValidationError
```

aiida_mlip/calculations/train.py
```diff
@@ -0,0 +1,170 @@
...
+        Error message if validation fails, None otherwise.
+    """
+    if "mlip_config" in port_namespace:
+        if "mlip_config" not in inputs:
+            raise InputValidationError("No config file given")
+        mlip_dict = inputs.mlip_config.as_dictionary()
+        list_of_keys = ["train_file", "valid_file", "test_file", "name"]
+        for key in list_of_keys:
```
ElliottKasoar, line 39:
```suggestion
        required_keys = ["train_file", "valid_file", "test_file", "name"]
        for key in required_keys:
```

aiida_mlip/calculations/train.py
```diff
@@ -0,0 +1,170 @@
...
+            if key not in mlip_dict:
+                raise InputValidationError(f"Mandatory key {key} not in config file")
+            # Check if the keys actually correspond to a path except name which is
+            # just the name to use for the output files
+            if key in mlip_dict and key != "name":
```
ElliottKasoar, line 44:
```suggestion
            if key != "name":
```
It will have raised an error `if key not in mlip_dict`

aiida_mlip/calculations/train.py
```diff
@@ -0,0 +1,170 @@
...
+                if Path(key).exists() is False:
```
ElliottKasoar, line 45:
```suggestion
                if not Path(key).exists():
```
(Not very different)

aiida_mlip/calculations/train.py
```diff
@@ -0,0 +1,170 @@
...
+        model_output = model_dir / f"{mlip_dict['name']}.model"
+        compiled_model_output = model_dir / f"{mlip_dict['name']}_compiled.model"
+
+        codeinfo = datastructures.CodeInfo()
+
+        # Initialize cmdline_params with train command
+        codeinfo.cmdline_params = ["train"]
+
+        for flag, value in cmd_line.items():
+            codeinfo.cmdline_params += [f"--{flag}", str(value)]
```
oerc0122, line 148: Given this is such a common pattern to all of these would it be best to turn this into a function?

aiida_mlip/parsers/train_parser.py
```diff
@@ -0,0 +1,117 @@
...
+        -------
+        int
+            An exit code.
+        """
+        remote_dir = Path(self.node.get_remote_workdir())
+        mlip_dict = self.node.get_option("mlip_config").as_dictionary()
+        log_dir = remote_dir / Path(mlip_dict.get("log_dir", "logs"))
+        checkpoint_dir = remote_dir / Path(
+            mlip_dict.get("checkpoint_dir", "checkpoints")
+        )
```
oerc0122, line 78: The right-hand argument doesn't need to be a `Path`, it can be a string and the `/` will handle the conversion

aiida_mlip/parsers/train_parser.py
```diff
@@ -0,0 +1,117 @@
...
+        results_dir = remote_dir / Path(mlip_dict.get("results_dir", "results"))
+        model_dir = remote_dir / Path(mlip_dict.get("model_dir", ""))
```
oerc0122, line 80: Looking at this, might it be better to have:
```suggestion
        remote_dirs = {typ: remote_dir / mlip_dict.get(f"{typ}_dir", default)
                       for typ, default in (("log", "logs"), 
                                            ("checkpoint", "checkpoints"),
                                            ("results", "results"),
                                            ("model", ""))}
```
And use the dict as the main reference for these.

aiida_mlip/parsers/train_parser.py
```diff
@@ -0,0 +1,117 @@
...
+
+        output_filename = self.node.get_option("output_filename")
+        model_output = model_dir / f"{mlip_dict['name']}.model"
+        compiled_model_output = model_dir / f"{mlip_dict['name']}_compiled.model"
+        result_name = results_dir / f"{mlip_dict['name']}_run-2024.txt"
```
oerc0122, line 85: Do we want "2024" in here?

oerc0122 (changes requested):

aiida_mlip/calculations/train.py
```diff
@@ -0,0 +1,170 @@
...
+            raise InputValidationError("No config file given")
+        mlip_dict = inputs.mlip_config.as_dictionary()
+        required_keys = ["train_file", "valid_file", "test_file", "name"]
+        for key in required_keys:
+            if key not in mlip_dict:
+                raise InputValidationError(f"Mandatory key {key} not in config file")
+            # Check if the keys actually correspond to a path except name which is
+            # just the name to use for the output files
+            if key != "name":
+                if not Path(key).exists():
```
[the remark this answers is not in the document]
federicazanca: yeah I'll do that I think

aiida_mlip/calculations/train.py
```diff
@@ -0,0 +1,170 @@
...
+            The calculation job process spec to define.
+        """
+        super().define(spec)
+        spec.inputs.validator = validate_inputs
+        # Define inputs
+        spec.input(
+            "mlip_config",
+            valid_type=JanusConfigfile,
+            required=False,
+            help="Mlip architecture to use for calculation, defaults to mace",
```
ElliottKasoar, line 93: This needs updating

aiida_mlip/calculations/train.py
```diff
@@ -0,0 +1,170 @@
...
+        )
+        spec.output("model", valid_type=ModelData)
+        spec.output("compiled_model", valid_type=SinglefileData)
+        spec.output(
+            "results_dict",
+            valid_type=Dict,
+            help="The `results_dict` output node of the successful calculation.",
```
ElliottKasoar, line 100:
```suggestion
            help="The `results_dict` output node of the training.",
```
?

(I'm not completely sure what goes into the output node, so there may well be a better description/it may be fine as is)

aiida_mlip/parsers/train_parser.py
```diff
@@ -0,0 +1,116 @@
...
+        Check that the ProcessNode being passed was produced by a `Singlepoint`.
+
+        Parameters
+        ----------
+        node : aiida.orm.nodes.process.process.ProcessNode
+            ProcessNode of calculation.
+        """
+        super().__init__(node)
+
+    # disable for now
```
ElliottKasoar, line 57: Is this still supposed to be here?

aiida_mlip/parsers/train_parser.py
ElliottKasoar: It comes from the [tag](https://github.com/ACEsuit/mace/blob/main/mace/cli/run_train.py#L58C17-L58C24) via the seed in the config (the default is [123](https://github.com/ACEsuit/mace/blob/main/mace/tools/arg_parser.py#L30)).

federicazanca: ah ok I missed this, I'll fix it. does the output names also use other things from the config? name, seed, others?

aiida_mlip/calculations/train.py
```diff
@@ -0,0 +1,197 @@
...
+        # The config file needs to be copied in the working folder
+        # Read content
+        mlip_dict = self.inputs.mlip_config.as_dictionary
+        config_parse = self.inputs.mlip_config.get_content()
+        # Extract paths from the config
+        abs_train_folder = (Path(mlip_dict["train_file"])).resolve()
+
+        abs_test_folder = (Path(mlip_dict["test_file"])).resolve()
+
+        abs_valid_folder = (Path(mlip_dict["valid_file"])).resolve()
```
oerc0122, line 149: Excess parens.
```suggestion
        abs_train_folder = Path(mlip_dict["train_file"]).resolve()

        abs_test_folder = Path(mlip_dict["test_file"]).resolve()

        abs_valid_folder = Path(mlip_dict["valid_file"]).resolve()
```

aiida_mlip/calculations/train.py
```diff
@@ -0,0 +1,197 @@
...
+        config_parse = config_parse.replace(
+            mlip_dict["train_file"], str(abs_train_folder)
+        )
+        config_parse = config_parse.replace(
+            mlip_dict["test_file"], str(abs_test_folder)
+        )
+        config_parse = config_parse.replace(
+            mlip_dict["valid_file"], str(abs_valid_folder)
+        )
+
```
oerc0122, line 160: Given the `abs_xxx` don't seem to be being reused (correct me if I'm wrong):
```suggestion
        for file in ("train_file", "test_file", "valid_file"):
            abs_path = Path(mlip_dict[file]).resolve()
            
            # Update the config file with absolute paths
            config_parse = config_parse.replace(
                mlip_dict[file], str(abs_folder)
            )
```

aiida_mlip/calculations/train.py
```diff
@@ -0,0 +1,197 @@
...
+        # Copy config file content inside the folder where the calculation is run
+        with folder.open("mlip_train.yml", "w", encoding="utf-8") as configfile:
```
oerc0122, line 162: Does this file need to be an input?

aiida_mlip/parsers/train_parser.py
```diff
@@ -0,0 +1,211 @@
...
+        Parameters
+        ----------
+        result_name : Path
+            Path to the result file.
+        """
+        with open(result_name, encoding="utf-8") as file:
+            last_dict_str = None
+            for line in file:
+                try:
+                    last_dict_str = ast.literal_eval(line.strip())
```
oerc0122, line 188: `ast.literal_eval` raises big concerns. What is this trying to do? What is the format of this file? If it's JSON like (Python `dict` like) should be using stdlib json parser.

tests/calculations/configs/mlip_train.yml
oerc0122, line 1: Do you need all these parameters or are some optional? Would it be worth marking the optional ones somehow?

oerc0122 (changes requested):

aiida_mlip/calculations/train.py
federicazanca: I think it's nicer to have it in the folder, like easier to find and have an immediate idea of what's in the folder where the calculation is run, but it doesn't need to I guess

tests/calculations/configs/mlip_train.yml
federicazanca: the only mandatory ones are those that I check in the validate_inputs. I can check in janus if it is indicated somewhere that the other parameters are optional, I try to do the same that is done there. or I can just write it in the documentation

aiida_mlip/calculations/train.py
oerc0122: Sorry, I meant does the name of `mlip_train.yml` need to be a variable? Sorry for being unclear. Writing in a hurry and knowing what I meant.

tests/calculations/configs/mlip_train.yml
oerc0122: So this is a Janus style input? Might just be worth adding `# Optional` to those which aren't mandatory to make it clear to users using this as an example that their input file doesn't need to be so complex.

aiida_mlip/calculations/train.py
federicazanca: do you mean if the user can change it and then it should be a variable that takes this into account? in this case no, this is just the name of a copied file that I want to be always the same.
if instead you mean to save it into a variable rather then passing the string every time I use it (which is twice in this code), I changed it now

aiida_mlip/calculations/train.py
```diff
@@ -139,25 +141,16 @@ def prepare_for_submission(
...
-        )
+        # Extract paths from the config
+        for file in ("train_file", "test_file", "valid_file"):
+            abs_path = Path(mlip_dict[file]).resolve()
 
+            # Update the config file with absolute paths
+            config_parse = config_parse.replace(mlip_dict[file], str(abs_path))
         # Copy config file content inside the folder where the calculation is run
-        with folder.open("mlip_train.yml", "w", encoding="utf-8") as configfile:
+        config_copy = "mlip_train.yml"
```
oerc0122, line 152: I actually meant whether this wants to be passed in as an argument in the config. But we can do that later.

federicazanca: there meaning in the other comment

tests/calculations/test_train.py
```diff
@@ -0,0 +1,98 @@
...
+    entry_point_name = "janus.train"
+    config_path = config_folder / "mlip_train.yml"
+
+    # Temporarily modify config file to introduce an error
+    with open(config_path, encoding="utf-8") as file:
+        right_path = file.read()
+
+    wrong_path = right_path.replace("mlip_train.xyz", "mlip_train_wrong.xyz")
+    with open(config_path, "w", encoding="utf-8") as file:
+        file.write(wrong_path)
```
oerc0122, line 55: I would feel more comfortable if we weren't modifying a file in-place. This seems liable to fall over and leave a corrupted file in case of an abort. 

Suggest dumping to a temporary file instead. Possibly using `pytest`'s tmp_path with multiple files generated on the fly.

federicazanca: fair enough, in general, would you say it' better to actually use a tmp_path or to just have different files saved and use those? because the reason I did this (and I would use a tmp_path for the same reason) is to not store many slightly different config files, so the folder looks tidier to me, but I could just do that, I don't know what way is considered better normally

oerc0122 (approved):

PR #128, opened 2024-05-23
add fine tuning

### LINKED ISSUE #127: add fine tuning

federicazanca:
Closes #127 
--fine_tune key added as an input for the train `Calcjob` 
The input validation checks that a model is given either in the config file or as an input.
in the config file, it can be anything (a string referring to some model types like "small" or a path to a model file)
If it is given as a `ModelData` input, it has to be a file. 
This is in general a main difference between janus and aiida-mlip that I don't know if we want to keep like this or change (the problem being that `ModelData` is built to be a file, so if we want it to also be a string, I can either completely modify it or just allow an additional input that is an alternative to the `ModelData` and it is just a string)

- [x] check that there aren't other checks to do 
- [x] more tests could be added where I actually run the calculation; I just wasn't sure because of the mace development thing
- [x] docs


tests/calculations/test_fine-tuning.py
```diff
@@ -0,0 +1,64 @@
...
+        calc_info.uuid,
+        "aiida-stdout.txt",
+        "logs",
+        "results",
+        "checkpoints",
+        "test.model",
+        "test_compiled.model",
+    ]
+    print(calc_info.codes_info[0].cmdline_params)
+    print(cmdline_params)
```
oerc0122, line 41: Debug prints?

oerc0122 (approved):
Looks good, just optional stylistic things

aiida_mlip/calculations/train.py
```diff
@@ -47,6 +47,15 @@ def validate_inputs(
             # Check if the keys actually correspond to a path
             if not ((Path(config_file.as_dictionary[key])).resolve()).exists():
                 raise InputValidationError(f"Path given for {key} does not exist")
+        # Check if fine-tuning is enabled and validate accordingly
+        if (
+            inputs["fine_tune"]
```
ElliottKasoar, line 52: Just to check, is `"fine_tune"` always a key at this point?

aiida_mlip/calculations/train.py
```diff
@@ -148,6 +169,13 @@ def prepare_for_submission(
 
             # Update the config file with absolute paths
             config_parse = config_parse.replace(mlip_dict[file], str(abs_path))
+
+        # Add foundation_model to the config file if fine-tuning is enabled
+        if self.inputs.fine_tune and "model" in self.inputs:
```
ElliottKasoar, line 174: Just to check, is `self.inputs.fine_tune` always defined here (I'm guessing yes if it's defined via the default `spec.input`)?

aiida_mlip/calculations/train.py
```diff
@@ -158,11 +186,16 @@ def prepare_for_submission(
...
-        cmd_line = {}
-        cmd_line["mlip-config"] = config_copy
+        cmd_line = {"mlip-config": config_copy}
+        if self.inputs.fine_tune:
+            cmd_line["fine-tune"] = None
+
         # Add cmd line params to codeinfo
         for flag, value in cmd_line.items():
-            codeinfo.cmdline_params += [f"--{flag}", str(value)]
+            if value is None:
```
ElliottKasoar, line 195: Might we ever want to actually pass None as a value?

Probably not since it'd end up as a "None" string, so probably just something to be aware of...

ElliottKasoar (commented):
Just a couple of questions/checks but looks good!

tests/calculations/test_train.py
```diff
@@ -97,3 +101,81 @@ def test_noname(
...
+    retrieve_list = [
+        calc_info.uuid,
+        "aiida-stdout.txt",
+        "logs",
+        "results",
+        "checkpoints",
+        "test.model",
+        "test_compiled.model",
+    ]
+    print(calc_info.codes_info[0].cmdline_params)
```
ElliottKasoar, line 134:
```suggestion
```

tests/calculations/test_train.py
```diff
@@ -97,3 +101,81 @@ def test_noname(
...
+        "metadata": {"options": {"resources": {"num_machines": 1}}},
+        "fine_tune": Bool(True),
+        "code": janus_code,
+        "mlip_config": config,
+        "model": ModelData.local_file(file=model_file, architecture="mace_mp"),
+    }
+
+    trainfinetuneCalc = CalculationFactory("janus.train")
+    result = run(trainfinetuneCalc, **inputs)
+    print(result["results_dict"].get_dict())
```
ElliottKasoar, line 177:
```suggestion
```

aiida_mlip/calculations/train.py
federicazanca: fine_tune can be True or False, and only if it's true we add {fine-tune:None} to the cmdline dict so that here it just adds `--fine-tune` to the command line  instead of writing like `janus --fine-tune True` it will just write `janus --fine-tune`)

aiida_mlip/calculations/train.py
federicazanca: it should since there is a default I think

alinelena (approved):

alinelena (approved):

alinelena (dismissed):

docs/source/user_guide/training.rst
```diff
@@ -76,6 +76,26 @@ while the other parameters are optional. Here is an example (can be found in the
     keep_isolated_atoms: True
     save_cpu: True
 
+It is also possible to fine-tune models using the same type of `Calcjob`.
+In that case some additional parameters must be used:
```
ElliottKasoar, line 80: It's probably worth saying explicitly which new parameters are required here, otherwise you have to check the difference between this and normal training (and assume only the minimum options have been listed in both cases).

aiida_mlip/calculations/train.py
```diff
@@ -89,6 +98,18 @@ def define(cls, spec: CalcJobProcessSpec) -> None:
...
+
+        spec.input(
+            "fine_tune",
+            valid_type=Bool,
+            required=False,
+            default=lambda: Bool(False),
+            help="Fine-tuning of a model",
+        )
+        spec.input(
+            "model", valid_type=ModelData, required=False, help="Model to fine-tune"
```
ElliottKasoar, line 110: I think this should be the [model type](https://github.com/ACEsuit/mace/blob/main/mace/tools/arg_parser.py#L92), which I don't think it unique to fine tuning

aiida_mlip/calculations/train.py
```diff
@@ -89,6 +98,18 @@ def define(cls, spec: CalcJobProcessSpec) -> None:
             required=True,
             help="Config file with parameters for training",
         )
+
+        spec.input(
+            "fine_tune",
+            valid_type=Bool,
+            required=False,
+            default=lambda: Bool(False),
+            help="Fine-tuning of a model",
```
ElliottKasoar, line 107:
```suggestion
            help="Whether fine-tuning a model",
```

aiida_mlip/calculations/train.py
federicazanca: is it also a keyword in the normal training?

docs/source/user_guide/training.rst
federicazanca: isn't the only additional thing the model?
 the parameters described here are all those that are possible to pass to aiida: the config file (which was also for the training and I understood it's more or less the same), the model (that can be given as in the example as an input or in the config, and that is for my understanding the only possible difference inside the config) and the fine-tune keyword to "activate" the fine tuning

federicazanca: (and then the code and metadata, which you pass to every calcjob)

aiida_mlip/calculations/train.py
ElliottKasoar: Yes, I think so.

This might be fine as is, depending on how it's used.

`model` in the yaml config that's passed to MACE via janus is the model type, but is this just used as the actual model itself, or is it supposed to be the value in the yaml?

(If we do eventually add options other than via the config, having different meanings could make things more complicated)

docs/source/user_guide/training.rst
ElliottKasoar: > isn't the only additional thing the model?

I think so, which was part of the point, as "some additional parameters" implies to me that there are more differences to worry about.

aiida_mlip/calculations/train.py
federicazanca: changed the keyword tofoundation_model

oerc0122 (approved):

docs/source/user_guide/training.rst
```diff
@@ -76,6 +76,33 @@ while the other parameters are optional. Here is an example (can be found in the
...
+    submit(TrainCalculation,inputs)
+
+A model to fine-tune has to be provided as an input, either as a `ModelData` type (in which case it has to be a model file), or in the config file at the keyword `foundation_model`.
+If the keyword `fine_tune` is True but no model is given either way, it will return an error.
+
+.. note::
+
+    The keyword 'model' and 'foundation_model' refer to two different things.
+    'foundation_model' is the path to the model to fine-tune (or a shortcut like 'small', etc).
+    'model' refers to the model-type (see `MACE <https://mace-docs.readthedocs.io/en/latest/>`_ documentation)
```
ElliottKasoar, line 103:
```suggestion
    'model' refers to the model-type (see `MACE <https://mace-docs.readthedocs.io/en/latest/guide/training.html#model>`_ documentation)
```

ElliottKasoar (approved):

PR #136, opened 2024-06-03
fix result dictionary

### LINKED ISSUE #135: fix results dictionary

federicazanca:
close #135

### DIFF
--- a/aiida_mlip/parsers/sp_parser.py
+++ b/aiida_mlip/parsers/sp_parser.py
@@ -98,7 +98,9 @@ class SPParser(BaseParser):
         with self.retrieved.open(xyz_output, "rb") as handle:
             self.out("xyz_output", SinglefileData(file=handle, filename=xyz_output))
 
-        content = read(Path(self.node.get_remote_workdir(), xyz_output))
+        content = read(
+            Path(self.node.get_remote_workdir(), xyz_output), format="extxyz"
+        )
         results = convert_numpy(content.todict())
         results_node = Dict(results)
         self.out("results_dict", results_node)
--- a/tests/calculations/configs/config_janus.yaml
+++ b/tests/calculations/configs/config_janus.yaml
@@ -1,4 +1,3 @@
 properties:
   - "energy"
-out: "results.xyz"
 arch: mace_mp
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,9 +28,9 @@ documentation = "https://stfc.github.io/aiida-mlip/"
 [tool.poetry.dependencies]
 python = "^3.9"
 aiida-core = "^2.5"
-ase = "^3.22.1"
+ase = "^3.23.0"
 voluptuous = "^0.14"
-janus-core = "^v0.5.0b0"
+janus-core = "^v0.6.0b0"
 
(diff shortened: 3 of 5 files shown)


aiida_mlip/parsers/sp_parser.py
```diff
@@ -98,7 +98,9 @@ def parse(self, **kwargs) -> int:
         with self.retrieved.open(xyz_output, "rb") as handle:
             self.out("xyz_output", SinglefileData(file=handle, filename=xyz_output))
 
-        content = read(Path(self.node.get_remote_workdir(), xyz_output))
+        content = read(
+            Path(self.node.get_remote_workdir(), xyz_output), format="extxyz"
```
ElliottKasoar, line 102: Why is this necessary?

In general janus allows any file format to be written out. Doesn't this limit that?

ElliottKasoar:
The test errors may well be related to ASE 3.23.0

ElliottKasoar (commented):
It might be nice to add a test that would fail before this change.

aiida_mlip/parsers/sp_parser.py
federicazanca: yeah I thought it would be only xyz files. 
but even if I didn't do this, I'm still reading it as a dictionary, if the output was another format I couldn't do that (? depends on the format I guess)

federicazanca:
> It might be nice to add a test that would fail before this change.



> It might be nice to add a test that would fail before this change.



> It might be nice to add a test that would fail before this change.



> It might be nice to add a test that would fail before this change.



> It might be nice to add a test that would fail before this change.

Yeah I'm trying to do this, I think actually maybe the failing test is because of this change (maybe)

aiida_mlip/parsers/sp_parser.py
ElliottKasoar: I'm not actually sure how many alternative file formats allow you to write out data from `info`/`arrays`/`calc.results`, but in general if it's stored, it should be readable again.

You can certainly get an `info` dictionary after reading a .cif file, although I'm not sure about how much can be written in the first place.

If it does only make sense to support extxyz, this should documented and probably (`--out`) validated.

federicazanca:
> The test errors may well be related to ASE 3.23.0

On second thought this is probably the cause of all the problems

tests/calculations/configs/config_janus.yaml
```diff
@@ -1,4 +1,3 @@
 properties:
   - "energy"
-out: "results.xyz"
```
ElliottKasoar, line 3: Does this not work, or do you just not want to test it anymore?

federicazanca: I removed it just to test something. But I think it should stay removed, cause the way the code is written at the moment it's always gonna use the default name for the output (which is what we want cause it's easier to handle for us and the user should not interact with the file directly anyways, so we know the output is always a xyz)

ElliottKasoar: So this isn't actually used at the moment anyway?

If you want to ensure the output is an xyz, you can also specify the format, which would also override any file extensions, but maybe that could be more confusing.

PR #138, opened 2024-06-04
134 model and architecture bug

### LINKED ISSUE #134: model and architecture bug
at the moment if model AND architecture are not provided anywhere the default is taken from janus. this is wrong, I just need to add a validation that at least one between arch or model should be given either as an input or in the config.
Then the thing is already written in a way that
if the architecture only is given , a default model is used (which is the same as janus indeed, but we handle it here in aiida level)
if the model only is given, the architecture should be already defined in the model. This actually is only true if the model is give through modeldata, if the model is given in the config, the architecture will be janus default, so probably we also want to change this.

federicazanca:
close #134 
I reorganised a little bit the base code for singlepoint, geometry and md to check that at least one between model and architecture is defined and then fixed some default values if one of the two is not defined. 
Before this the used default were from janus but we want to have our own.


aiida_mlip/calculations/base.py
```diff
@@ -44,6 +46,17 @@ def validate_inputs(
...
+    if (
+        "arch" not in inputs
+        and "model" not in inputs
+        and (
+            "config" not in inputs
+            or ("arch" not in inputs["config"] and "model" not in inputs["config"])
+        )
+    ):
+        raise InputValidationError(
+            "Either 'arch' or 'model' must be specified in the inputs or config"
```
ElliottKasoar, line 58:
```suggestion
            "'arch' and 'model' must both be specified in the inputs or config"
```
Otherwise it reads as one or the other, to me

tests/calculations/test_singlepoint.py
```diff
@@ -107,6 +110,24 @@ def test_sp_nostruct(fixture_sandbox, generate_calc_job, model_folder, janus_cod
         generate_calc_job(fixture_sandbox, entry_point_name, inputs)
 
 
+def test_sp_nomodel(fixture_sandbox, generate_calc_job, config_folder, janus_code):
+    """Test singlepoint calculation with error input"""
+    entry_point_name = "mlip.sp"
+
+    nacl = bulk("NaCl", "rocksalt", a=5.63)
+    write("NaCl.cif", nacl)
```
ElliottKasoar, line 118: Given that this isn't testing the config reader, I'd consider defining struct (and writing here) using temporary files, which would mean you don't need to unlink.

aiida_mlip/calculations/base.py
```diff
@@ -274,3 +255,77 @@ def prepare_for_submission(
         ]
 
         return calcinfo
+
+    def _define_architecture(self, cmd_line: dict) -> Union[dict, str]:
```
ElliottKasoar, line 259:
```suggestion
    def _define_architecture(self, cmd_line: dict) -> tuple[dict, str]:
```

aiida_mlip/calculations/base.py
```diff
@@ -274,3 +255,77 @@ def prepare_for_submission(
...
+        ----------
+        cmd_line : dict
+            Dictionary containing the cmd line keys.
+
+        Returns
+        -------
+        cmd_line_updated: dict
+            Dictionary containing the cmd line keys updated with the architecture.
+        architecture: str
+            Architecture type, either from inputs/config or a default value ("mace_mp").
```
ElliottKasoar, line 273:
```suggestion
        tuple[dict, str]
            Dictionary containing the cmd line keys updated with the architecture, and
            architecture type, either from inputs/config or a default value ("mace_mp").
```

aiida_mlip/calculations/base.py
```diff
@@ -274,3 +255,77 @@ def prepare_for_submission(
...
+        ----------
+        cmd_line : dict
+            Dictionary containing the cmd line keys.
+        architecture : str
+            Architecture type.
+
+        Returns
+        -------
+            dict
+                Dictionary containing the cmd line keys updated with the model.
```
ElliottKasoar, line 307:
```suggestion
        dict
            Dictionary containing the cmd line keys updated with the model.
```

aiida_mlip/calculations/base.py
```diff
@@ -274,3 +255,77 @@ def prepare_for_submission(
...
+
+        Returns
+        -------
+        cmd_line_updated: dict
+            Dictionary containing the cmd line keys updated with the architecture.
+        architecture: str
+            Architecture type, either from inputs/config or a default value ("mace_mp").
+        """
+        architecture = None
+        cmd_line_updated = cmd_line
```
ElliottKasoar, line 276: Are you wanting this to be a copy of `cmd_line`? 

At the moment, `cmd_line_updated is cmd_line == True` - they refer to the same object.

This means when you change `cmd_line_updated` after this, you change `cmd_line`, both in this function, and outside too (so you don't actually have to return the dictionary at all).

(This is also relevant in the other new function.)

aiida_mlip/calculations/base.py
federicazanca: I did mean what I wrote but we do actually want what you wrote (which might mean also I need to change something in the rest of the code)

aiida_mlip/calculations/base.py
federicazanca: I was trying to make a copy yes (it should be cmd_line.copy() instead right?)
But I'm gonna change cmd_line anyway, so it might as well not be a copy... but I understood it is better to make copies from previous prs in which I was doing something similar?

aiida_mlip/calculations/base.py
ElliottKasoar: Sorry you're right, I didn't read the logic carefully enough. I guess it depends how/where you treat defaults etc.

alinelena:
will need rebase after 137 gets in

aiida_mlip/calculations/base.py
oerc0122: Yes, it should be `cmd_line.copy()` for a shallow copy.

aiida_mlip/calculations/base.py
```diff
@@ -274,3 +253,76 @@ def prepare_for_submission(
...
+        """
+        model_path = None
+        cmd_line_updated = cmd_line
+        if "model" in self.inputs:
+            # Raise error if model is None
+            if self.inputs.model is None:
+                raise ValueError("Model cannot be None")
+            model_path = self.inputs.model.filepath
+        else:
+            if "config" in self.inputs and "model" in self.inputs.config:
```
oerc0122, line 314: Could be flattened.
```suggestion
        elif "config" in self.inputs and "model" in self.inputs.config:
```

tests/calculations/test_singlepoint.py
oerc0122: To do this you can either use `tempfile.NamedTemporaryFile` or use the `tmp_path` fixture from pytest (pass `tmp_path` as an arugment) and put the written file in `tmp_path / "NaCl.cif"`

oerc0122 (commented):
Just supporting @ElliottKasoar 's points.

tests/calculations/test_singlepoint.py
federicazanca: I just changed it in a way that I don't actually need a file

aiida_mlip/calculations/base.py
```diff
@@ -44,6 +44,27 @@ def validate_inputs(
...
+
+    if "model" not in inputs and (
+        "config" not in inputs or "model" not in inputs["config"]
+    ):
+        raise InputValidationError(
+            "'model' must be specified either in the inputs or in the config file"
+        )
+
+    if "arch" in inputs and "model" in inputs:
+        if inputs["arch"].value is not inputs["model"].architecture:
```
oerc0122, line 64: Can be flattened. 
```suggestion
    if "arch" in inputs and "model" in inputs and inputs["arch"].value is not inputs["model"].architecture:
```

aiida_mlip/calculations/base.py
```diff
@@ -206,42 +227,10 @@ def prepare_for_submission(
...
-                if "arch" in self.inputs:
-                    # if model is not given (which is different than it being None)
-                    model_path = ModelData.download(
-                        "https://github.com/stfc/janus-core/raw/main/tests/models/mace_mp_small.model",  # pylint: disable=line-too-long
-                        architecture,
-                    ).filepath
-        if model_path:
-            cmd_line.setdefault("calc-kwargs", {})["model"] = model_path
+        cmd_line = self._add_arch_to_cmdline(cmd_line)
+        cmd_line = self._add_model_to_cmdline(cmd_line)
```
oerc0122, line 231: If these are only ever returning back to `cmd_line`, could modify it in-place in the function.

aiida_mlip/calculations/base.py
```diff
@@ -274,3 +263,62 @@ def prepare_for_submission(
...
+        if "model" in self.inputs and hasattr(self.inputs.model, "architecture"):
+            architecture = str((self.inputs.model).architecture)
+            cmd_line_updated["arch"] = architecture
+        elif "arch" in self.inputs:
+            architecture = str(self.inputs.arch.value)
+            cmd_line_updated["arch"] = architecture
+        # At this point we must have the arch in the config so we don't need to write
+        # in the cmd line if in config
+        elif "config" in self.inputs and "arch" in self.inputs.config:
+            architecture = self.inputs.config.as_dictionary["arch"]
```
oerc0122, line 292: Is `architecture` being used here?

aiida_mlip/calculations/base.py
```diff
@@ -206,42 +226,10 @@ def prepare_for_submission(
...
-                if "arch" in self.inputs:
-                    # if model is not given (which is different than it being None)
-                    model_path = ModelData.download(
-                        "https://github.com/stfc/janus-core/raw/main/tests/models/mace_mp_small.model",  # pylint: disable=line-too-long
-                        architecture,
-                    ).filepath
-        if model_path:
-            cmd_line.setdefault("calc-kwargs", {})["model"] = model_path
+        cmd_line = self._add_arch_to_cmdline(cmd_line)
+        cmd_line = self._add_model_to_cmdline(cmd_line)
```
alinelena, line 230: i am a little bit confused here... the model seems to overwrite the arch... should not be additive?

aiida_mlip/calculations/base.py
```diff
@@ -319,5 +316,5 @@ def _add_model_to_cmdline(
             # No need for command line
             model_path = None
         if model_path:
-            cmd_line_updated.setdefault("calc-kwargs", {})["model"] = model_path
-        return cmd_line_updated
+            cmd_line.setdefault("calc-kwargs", {})["model"] = model_path
+        return cmd_line
```
oerc0122, line 320: For symmetry, does this one actually want to `return`?

oerc0122 (changes requested):

tests/calculations/test_singlepoint.py
```diff
@@ -61,48 +62,66 @@ def test_singlepoint(fixture_sandbox, generate_calc_job, janus_code, model_folde
...
-
+def test_sp_nostruct(fixture_sandbox, generate_calc_job, model_folder, janus_code):
+    """Test singlepoint calculation with error input"""
     entry_point_name = "mlip.sp"
+    model_file = model_folder / "mace_mp_small.model"
+    # pylint:disable=line-too-long
     inputs = {
         "metadata": {"options": {"resources": {"num_machines": 1}}},
         "code": janus_code,
         "arch": Str("mace"),
```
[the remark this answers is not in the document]
federicazanca: it has to do with what data you want to save (like data you see in the provenance graph).
Also I'm not sure if the thing is by default that all the new inputs have to be aiida nodes, and just the metadata is different

aiida_mlip/calculations/base.py
```diff
@@ -274,3 +266,58 @@ def prepare_for_submission(
...
+            Dictionary containing the cmd line keys updated with the model.
+        """
+        model_path = None
+        if "model" in self.inputs:
+            # Raise error if model is None (different than model not given as input)
+            if self.inputs.model is None:
+                raise ValueError("Model cannot be None")
+            model_path = self.inputs.model.filepath
+        elif "config" in self.inputs and "model" in self.inputs.config:
+            # No need for command line
```
oerc0122, line 320: As above

aiida_mlip/calculations/base.py
```diff
@@ -274,3 +266,58 @@ def prepare_for_submission(
...
+        """
+        architecture = None
+        if "model" in self.inputs and hasattr(self.inputs.model, "architecture"):
+            architecture = str((self.inputs.model).architecture)
+        elif "arch" in self.inputs:
+            architecture = str(self.inputs.arch.value)
+        # At this point we must have the arch in the config so we don't need to write
+        # in the cmd line if in config
+        elif "config" in self.inputs and "arch" in self.inputs.config:
+            architecture = None
```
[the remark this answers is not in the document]
federicazanca: no, it was for me to have the logic clearer but can remove it

oerc0122 (approved):

Issue #146, opened 2024-07-05
add high-throughput screening workchain


federicazanca:
Adding a workchain that takes files from a folder and runs a given calculation on all the files. The use for this is to use it later inside a training workchain.
Problems:
- [ ] Janus-core and aiida-qe are incompatible see Issue 196 in janus
[https://github.com/stfc/janus-core/issues/196](url)
- [ ] Adding more codes with different entry points might make the thing slightly more complex than expected. Basically I would like to give as an input an entry point for the calculations to run and then use it to set up a calculation. 
To explain better: this is how you set up a calcjob or a workflow. Let's use the entry points mlip.opt and quantumespresso.pw.relax
```
geomopt_janus = CalculationFactory("mlip.opt")
geomopt_qe = WorkflowFactory("quantumespresso.pw.relax")
```
Then this is how the WorkChain takes the inputs. Basically spec.input are inputs for the first steps of the workchain, and then every subprocess takes its own inputs in the form of a dictionary with the expose_inputs function.
```
class HTSWorkChain(WorkChain):
  @classmethod
   def define(cls, spec):
        super().define(spec)

        spec.input("folder", valid_type=Str, help="Folder containing CIF files")

        # spec.input("entrypoint", valid_type=Str, help="calculation entry point")

        spec.expose_inputs(geomopt_janus, namespace="janus_inputs", exclude="struct", required=False)
        spec.expose_inputs(geomopt_qe, namespace="qe_inputs", exclude="struct", required=False)
```
At the moment I cannot use the "entrypoint" input to chose the type of calculation to do inside the workchain, because both the input and expose_input need to be defined outside, so I have to have basically different keywords for any calculation I want to run high-throughputly (in this example I want to run either the janus opt or the qe-workchain for opt).
I will ask about this in the aiida.discourse, but if this is not fixable basically I will need to manually choose which types of calculations I want to add to the workchain.

**TLDR
making a high-throughput screening for different types of calculations (quantumespresso, janus opt, janus sp etc) might make it uselessly complex for me and the user, so I need to decide what's worth doing: 
what calculations do we want in the high-throughput screening workchain? are janus optimisation and qe optimisation enough?
Should I just not make a high-throughput workchain and proceed with the training?**

PR #147, opened 2024-07-05
High throughput wc

federicazanca:
Still very messy, just making the PR so we can actually see the code when discussing.

Ignore the comments in the model file, will remove and use them to fix other issues.

I made a `WorkChain` for submitting high-throughput screening, but I am facing the problems described in #146 

I also added the submission script for submitting the actual `Workchain` (in examples/workflows)

And a script to run high-throughput screening without writing a `WorkChain`. We might end up only keeping this? Depending on discussion on how to implement some things


pyproject.toml
```diff
@@ -79,6 +79,10 @@ build-backend = "poetry.core.masonry.api"
 "mlip.md_parser" = "aiida_mlip.parsers.md_parser:MDParser"
 "mlip.train_parser" = "aiida_mlip.parsers.train_parser:TrainParser"
 
+[tool.poetry.plugins."aiida.workflows"]
+"mlip.hts" = "aiida_mlip.workflows.hts:HTSWorkChain"
+"mlip.hts_wg" = "aiida_mlip.workflows.hts_workgraph:HTSWorkGraph"
```
ElliottKasoar, line 84: Are these both needed? I only see `mlip.hts` referenced anywhere.

examples/workflows/html/hts_workflow.html
ElliottKasoar, line 1: Is this supposed to be here?

examples/workflows/run_hts_
ElliottKasoar, line 1: Is this supposed to be here?

aiida_mlip/workflows/hts_workgraph.py
```diff
@@ -0,0 +1,64 @@
+""" Workgraph to run DFT calculations and use the outputs fpr training a MLIP model."""
```
ElliottKasoar, line 1:
```suggestion
"""WorkGraph to run DFT calculations and use the outputs for training a MLIP model."""
```
We probably want a file name that distinguishes this from hts.py better too

federicazanca:
UPDATE
everything changed since the beginning.
The high throughput workchain is now a `WorkGraph` instead. The `WorkChain` is cancelled.
I also put scripts in the example to submit it and also to submit ust many calculations without the need of the workgraph or workchain. Both options are valid and depend on the need of the user. 
The tests fail (at least in my computer) probably because of some problem in submitting to a queue. --> discuss with aiida people/fix after cloud is set up

aiida_mlip/workflows/hts_workgraph.py
```diff
@@ -0,0 +1,75 @@
...
+        except Exception:  # pylint: disable=broad-except
+            continue
+        structure = load_structure(child)
+        janus_opt_inputs["struct"] = structure
+        relax = wg.add_task(Geomopt, name=f"relax_{child.stem}", **janus_opt_inputs)
+        relax.set_context({"final_structure": f"relax.{child.stem}"})
+    return wg
+
+
+def HTSWorkGraph(folder_path: Path, inputs: dict) -> WorkGraph:
```
ElliottKasoar, line 45: Probably should be `create_hts_workgraph` or something else more in the style of a function name?

Also, is this specific to running geomopt, or can it be generalised to other calculations? Assuming it's specific, which I think it is, it's quite confusing.

aiida_mlip/workflows/hts_workgraph.py
```diff
@@ -0,0 +1,75 @@
...
+
+    Returns
+    -------
+    WorkGraph
+        The workgraph containing the high-throughput workflow.
+    """
+    wg = WorkGraph("hts_workflow")
+
+    wg.add_task(
+        run_opt_calc, name="opt_task", folder=folder_path, janus_opt_inputs=inputs
```
ElliottKasoar, line 64: Could this be generalised to take an arbitrary task using one of the function's parameters, so it wouldn't have to be hardcoded?

examples/workflows/submit_hts_workgraph.py
```diff
@@ -0,0 +1,20 @@
...
+from aiida.orm import load_code
+
+from aiida_mlip.data.model import ModelData
+from aiida_mlip.workflows.hts_workgraph import HTSWorkGraph
+
+folder_path = Path("/home/federica/aiida-mlip/tests/workflows/structures/")
+inputs = {
+    "model": ModelData.from_local(
+        "/home/federica/aiida-mlip/tests/data/input_files/mace/mace_mp_small.model",
+        architecture="mace_mp",
```
ElliottKasoar, line 14: Paths should be generalised

pyproject.toml
```diff
@@ -30,7 +30,8 @@ python = "^3.9"
 aiida-core = "^2.6"
 ase = "^3.23.0"
 voluptuous = "^0.14"
-janus-core = "^v0.6.0b0"
+janus-core = "^v0.6.3b0"
```
ElliottKasoar, line 33: Might need to pin or check compatibility with newer versions

tests/workflows/test_hts.py
```diff
@@ -0,0 +1,25 @@
...
+    }
+    wg = HTSWorkGraph(folder_path=structure_folder2, inputs=inputs)
+    wg.wait(15)
+
+    # AT THE MOMENT WE ONLY CHECK THE PROCESS IS CREATED AT LEAST,
+    #  WHEN WE FIX THE SUBMISSION THIS NEEDS TO BE CHANGED
+
+    assert wg.state == "CREATED"
+    # wg_node = load_node(wg.pk)
+    # assert isinstance(wg_node.outputs.opt_structures.h2o, StructureData)
```
ElliottKasoar, line 25: This needs updating

tests/workflows/test_hts.py
```diff
@@ -0,0 +1,25 @@
...
+def test_hts_wg(janus_code, structure_folder2, model_folder) -> None:
+    """Submit simple calcjob."""
+    model_file = model_folder / "mace_mp_small.model"
+    inputs = {
+        "model": ModelData.from_local(model_file, architecture="mace"),
+        "metadata": {"options": {"resources": {"num_machines": 1}}},
+        "code": janus_code,
+    }
+    wg = HTSWorkGraph(folder_path=structure_folder2, inputs=inputs)
+    wg.wait(15)
```
ElliottKasoar, line 18: Are there nicer ways of doing this?

ElliottKasoar (changes requested):

aiida_mlip/workflows/hts_workgraph.py
```diff
@@ -0,0 +1,75 @@
...
+    )
+
+    wg.group_outputs = [{"name": "opt_structures", "from": "opt_task.final_structures"}]
+
+    wg.to_html()
+
+    wg.max_number_jobs = 10
```
superstar54, line 71: Two small suggestions:
- set `max_number_jobs` as an input parameter.
- Not ot submit the wg inside the function, but return the `wg`, and let the user run `wg.submit()`. In this case, user can also modify the inputs of the tasks or other settings before submission.

aiida_mlip/workflows/hts_workgraph.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+    wg = WorkGraph("hts_workflow")
+
+    wg.add_task(
+        run_opt_calc, name="opt_task", folder=folder_path, janus_opt_inputs=inputs
+    )
+
+    wg.group_outputs = [{"name": "opt_structures", "from": "opt_task.final_structures"}]
+
+    wg.to_html()
```
superstar54, line 69: will this work in the notebook? if not, maybe delete this line.
```suggestion

```

pyproject.toml
```diff
@@ -79,6 +80,9 @@ build-backend = "poetry.core.masonry.api"
 "mlip.md_parser" = "aiida_mlip.parsers.md_parser:MDParser"
 "mlip.train_parser" = "aiida_mlip.parsers.train_parser:TrainParser"
 
+[tool.poetry.plugins."aiida.workflows"]
+"mlip.hts_wg" = "aiida_mlip.workflows.hts_workgraph:HTSWorkGraph"
+
```
superstar54, line 85: We are working on a new entry point for the WorkGraph. Will let you know the syntax when the API is fixed.

superstar54 (commented):
Hi! Thanks for the nice work and investigating the aiida-workgraph. I added a few small comments.

PR #149, opened 2024-07-08
change aiida version

federicazanca:
do we want to do this or just leave it as it was? 
it's the latest aiida aupdate with some stuff we discussed today in the meeting


federicazanca:
> (You may need to tell the docs to ignore the failed class imports via `nitpick_ignore` in `conf.py`)

yeah I was wondering about that

ElliottKasoar (approved):

PR #150, opened 2024-07-10
142 model download change

### LINKED ISSUE #142: Model download change
Following discussion with Giovanni. 
Probably should change the download to not save the downloaded model also in a local folder (which is what I do now) but only in the database.
Then fix the references to the filepath throughout the code: at the moment I use the filepath of the downloaded model sometimes to do stuff, which is obviously not shareable. The only thing I did to address this is to write a note in the documentation about it. If I change it I can remove

federicazanca:
Close #142 
Changed things:
- [x] When downloaded the model file is not saved as a file, just in the database unless the keyword keep_file is used
- [x] The existence of the file is checked via "model_hash"
- [x] Changed names of the functions to follow convention
- [x] Removed filepath attribute, which made it more difficult to share data since it referred to a file path in a local computer. And made some changes in the code related to this removal

### DIFF
--- a/aiida_mlip/calculations/base.py
+++ b/aiida_mlip/calculations/base.py
@@ -1,5 +1,7 @@
 """Base class for features common to most calculations."""
 
+import shutil
+
 from ase.io import read, write
 
 from aiida.common import InputValidationError, datastructures
@@ -199,8 +201,6 @@ class BaseJanus(CalcJob):  # numpydoc ignore=PR01
             An instance of `aiida.common.datastructures.CalcInfo`.
         """
 
-        # Create needed inputs
-
         if "struct" in self.inputs:
             structure = self.inputs.struct
         elif "config" in self.inputs and "struct" in self.inputs.config.as_dictionary:
@@ -211,8 +211,8 @@ class BaseJanus(CalcJob):  # numpydoc ignore=PR01
         # Transform the structure data in xyz file called input_filename
         input_filename = self.inputs.metadata.options.input_filename
         atoms = structure.get_ase()
-        # with folder.open(input_filename, mode="w", encoding='utf8') as file:
-        write(folder.abspath + "/" + input_filename, images=atoms)
+        with folder.open(input_filename, mode="w", encoding=None) as file:
+            write(file.name, images=atoms)
 
         log_filename = (self.inputs.log_filename).value
         cmd_line = {
@@ -231,7 +231,7 @@ class BaseJanus(CalcJob):  # numpydoc ignore=PR01
         # Define architecture from model if model is given,
         # otherwise get architecture from inputs and download default model
         self._add_arch_to_cmdline(cmd_line)
-        self._add_model_to_cmdline(cmd_line)
+        self._add_model_to_cmdline(cmd_line, folder)
 
         if "config" in self.inputs:
             # Add config file to command line
@@ -290,8 +290,7 @@ class BaseJanus(CalcJob):  # numpydoc ignore=PR01
             cmd_line["arch"] = architecture
 
     def _add_model_to_cmdline(
-        self,
-        cmd_line: dict,
+        self, cmd_line: dict, folder: aiida.common.folders.Folder
     ) -> dict:
         """
         Find model in inputs or config file and add to command line if needed.
@@ -300,6 +299,9 @@ class BaseJanus(CalcJob):  # numpydoc ignore=PR01
         ----------
         cmd_line : dict
             Dictionary containing the cmd line keys.
+
+        folder : aiida.common.folders.Folder
+            An `aiida.common.folders.Folder` to temporarily write files on disk.
 
         Returns
         -------
@@ -311,6 +313,11 @@ class BaseJanus(CalcJob):  # numpydoc ignore=PR01
             # Raise error if model is None (different than model not given as input)
             if self.inputs.model is None:
                 raise ValueError("Model cannot be None")
-            model_path = self.inputs.model.filepath
+
+            with self.inputs.model.open(mode="rb") as source:
+                with folder.open("modelcopy.model", mode="wb") as target:
+                    shutil.copyfileobj(source, target)
+
+            model_path = "modelcopy.model"
         if model_path:
             cmd_line.setdefault("calc-kwargs", {})["model"] = model_path
--- a/aiida_mlip/calculations/base.py
+++ b/aiida_mlip/calculations/base.py
@@ -211,7 +211,7 @@ class BaseJanus(CalcJob):  # numpydoc ignore=PR01
         # Transform the structure data in xyz file called input_filename
         input_filename = self.inputs.metadata.options.input_filename
         atoms = structure.get_ase()
-        with folder.open(input_filename, mode="w", encoding=None) as file:
+        with folder.open(input_filename, mode="w", encoding="utf8") as file:
             write(file.name, images=atoms)
 
         log_filename = (self.inputs.log_filename).value
@@ -319,5 +319,4 @@ class BaseJanus(CalcJob):  # numpydoc ignore=PR01
                     shutil.copyfileobj(source, target)
 
             model_path = "modelcopy.model"
-        if model_path:
             cmd_line.setdefault("calc-kwargs", {})["model"] = model_path
--- a/aiida_mlip/calculations/base.py
+++ b/aiida_mlip/calculations/base.py
@@ -65,10 +65,13 @@ def validate_inputs(
     if (
         "arch" in inputs
         and "model" in inputs
-        and inputs["arch"].value is not inputs["model"].architecture
+        and inputs["arch"].value != inputs["model"].architecture
     ):
+        inputvalue = inputs["arch"].value
+        modelvalue = inputs["model"].architecture
         raise InputValidationError(
-            "'arch' in ModelData and in 'arch' input must be the same"
+            "'arch' in ModelData and in inputs must be the same, "
+            f"but they are {modelvalue} and {inputvalue}"
         )
 
 
--- a/aiida_mlip/calculations/base.py
+++ b/aiida_mlip/calculations/base.py
@@ -300,7 +300,7 @@ class BaseJanus(CalcJob):  # numpydoc ignore=PR01
         cmd_line : dict
             Dictionary containing the cmd line keys.
 
-        folder : aiida.common.folders.Folder
+        folder : ~aiida.common.folders.Folder
             An `aiida.common.folders.Folder` to temporarily write files on disk.
 
         Returns
--- a/aiida_mlip/calculations/base.py
+++ b/aiida_mlip/calculations/base.py
@@ -317,9 +317,11 @@ class BaseJanus(CalcJob):  # numpydoc ignore=PR01
             if self.inputs.model is None:
                 raise ValueError("Model cannot be None")
 
-            with (self.inputs.model.open(mode="rb") as source,
-                  folder.open("mlff.model", mode="wb") as target):
-                    shutil.copyfileobj(source, target)
+            with (
+                self.inputs.model.open(mode="rb") as source,
+                folder.open("mlff.model", mode="wb") as target,
+            ):
+                shutil.copyfileobj(source, target)
 
             model_path = "mlff.model"
             cmd_line.setdefault("calc-kwargs", {})["model"] = model_path
--- a/aiida_mlip/calculations/base.py
+++ b/aiida_mlip/calculations/base.py
@@ -315,8 +315,8 @@ class BaseJanus(CalcJob):  # numpydoc ignore=PR01
                 raise ValueError("Model cannot be None")
 
             with self.inputs.model.open(mode="rb") as source:
-                with folder.open("modelcopy.model", mode="wb") as target:
+                with folder.open("mlff.model", mode="wb") as target:
                     shutil.copyfileobj(source, target)
 
-            model_path = "modelcopy.model"
+            model_path = "mlff.model"
             cmd_line.setdefault("calc-kwargs", {})["model"] = model_path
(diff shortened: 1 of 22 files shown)


federicazanca:
Note: there is the same issue with the filepath in the config file, but actually I ended up  never using filepath in the calcjobs or anywhere else, so it's a problem mostly cause it's useless lines of code, but it's usable without issues I think. There are in general  some things to change in the config file cause I ended up not using some features, but it's not too urgent.

aiida_mlip/calculations/base.py
```diff
@@ -211,8 +211,8 @@ def prepare_for_submission(
         # Transform the structure data in xyz file called input_filename
         input_filename = self.inputs.metadata.options.input_filename
         atoms = structure.get_ase()
-        # with folder.open(input_filename, mode="w", encoding='utf8') as file:
-        write(folder.abspath + "/" + input_filename, images=atoms)
+        with folder.open(input_filename, mode="w", encoding=None) as file:
```
ElliottKasoar, line 214: Why `None`?

aiida_mlip/calculations/base.py
```diff
@@ -311,6 +313,11 @@ def _add_model_to_cmdline(
...
             if self.inputs.model is None:
                 raise ValueError("Model cannot be None")
-            model_path = self.inputs.model.filepath
+
+            with self.inputs.model.open(mode="rb") as source:
+                with folder.open("modelcopy.model", mode="wb") as target:
+                    shutil.copyfileobj(source, target)
+
+            model_path = "modelcopy.model"
         if model_path:
```
ElliottKasoar, line 322: Does this not now only depend on `if "model" in self.inputs`?

aiida_mlip/calculations/base.py
```diff
@@ -311,6 +313,11 @@ def _add_model_to_cmdline(
             # Raise error if model is None (different than model not given as input)
             if self.inputs.model is None:
                 raise ValueError("Model cannot be None")
-            model_path = self.inputs.model.filepath
+
+            with self.inputs.model.open(mode="rb") as source:
+                with folder.open("modelcopy.model", mode="wb") as target:
```
ElliottKasoar, line 318: So is `"modelcopy.model`" a temporary copy of the model? Why is it needed and referenced in the tests?

ElliottKasoar (commented):
Looks good so far.

A few suggestions for tidying up, and a few more general questions where comments here and/or in the code itself my be useful

aiida_mlip/data/model.py
```diff
@@ -231,32 +191,37 @@ def download(
...
+                if i["ModelData_1"]["ctime"] != model.ctime:
+                    delete_nodes(
+                        [model.uuid],
+                        dry_run=False,
+                        create_forward=True,
+                        call_calc_forward=True,
+                        call_work_forward=True,
+                    )
+                    model = load_node(i["ModelData_1"]["pk"])
+                    break
```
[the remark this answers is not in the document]
federicazanca: yes this is what it is doing, it's checking if in the database there is already a model with the same hash. ctime is creation time yes, I'm just checking if the model I found is the one I just created. 
This is what the querybuilder returns:
```
In [4]: qb = QueryBuilder()

In [5]: qb.append(ModelData,project=["attributes","pk","ctime"])
Out[5]: QueryBuilder(path=[{'entity_type': 'data.mlip.modeldata.ModelData.', 'orm_base': 'node', 'tag': 'ModelData_1', 'joining_keyword': None, 'joining_value': None, 'edge_tag': None, 'outerjoin': False}], filters={'ModelData_1': {'node_type': {'like': 'data.mlip.modeldata.%'}}}, project={'ModelData_1': [{'attributes': {}}, {'pk': {}}, {'ctime': {}}]}, project_map={}, order_by=[], limit=None, offset=None, distinct=False)

In [6]: qb.dict()
Out[6]: 
[{'ModelData_1': {'attributes': {'filename': 'small.model',
    'filepath': '/home/federica/dtk/aiida_interactions/models/mace_mp/small.model',
    'architecture': 'mace_mp'},
   'pk': 4834,
   'ctime': datetime.datetime(2024, 6, 12, 16, 31, 33, 965922, tzinfo=datetime.timezone(datetime.timedelta(seconds=3600)))}},
 {'ModelData_1': {'attributes': {'filename': 'small.model',
    'filepath': '/home/federica/dtk/aiida_interactions/models/mace_mp/small.model',
    'architecture': 'mace_mp'},
   'pk': 4948,
   'ctime': datetime.datetime(2024, 6, 12, 17, 13, 35, 23590, tzinfo=datetime.timezone(datetime.timedelta(seconds=3600)))}},
```
I don't know why they are called ModelData_1, prob can change it

aiida_mlip/calculations/base.py
federicazanca: Probably can put it back to utf8, I was trying things and thought that could be a problem but prob not

aiida_mlip/calculations/base.py
federicazanca: it's the file you will actually have in the folder where the calc is actually run. could have a better name

tests/data/test_model.py
```diff
@@ -30,27 +30,28 @@ def test_relativepath():
...
         filename="model",
         architecture="mace",
     )
     assert model.architecture == "mace"
 
 
-def test_download_fresh_file(tmp_path):
-    """Test if download works"""
+def test_download_fresh_file_keep(tmp_path):
+    """Test if download works and the downloaded file is kept in them chosen folder"""
```
ElliottKasoar, line 42:
```suggestion
    """Test if download works and the downloaded file is kept in the chosen folder"""
```

aiida_mlip/calculations/base.py
```diff
@@ -311,6 +313,10 @@ def _add_model_to_cmdline(
             # Raise error if model is None (different than model not given as input)
             if self.inputs.model is None:
                 raise ValueError("Model cannot be None")
-            model_path = self.inputs.model.filepath
-        if model_path:
+
+            with self.inputs.model.open(mode="rb") as source:
+                with folder.open("modelcopy.model", mode="wb") as target:
```
alinelena, line 318:
```suggestion
                with folder.open("mlff.model", mode="wb") as target:
```

aiida_mlip/calculations/base.py
```diff
@@ -311,6 +313,10 @@ def _add_model_to_cmdline(
...
+                    shutil.copyfileobj(source, target)
+
+            model_path = "modelcopy.model"
```
alinelena, line 321:
```suggestion
            model_path = "mlff.model"
```

aiida_mlip/calculations/train.py
```diff
@@ -175,9 +176,10 @@ def prepare_for_submission(
 
         # Add foundation_model to the config file if fine-tuning is enabled
         if self.inputs.fine_tune and "foundation_model" in self.inputs:
-            model_data = self.inputs.foundation_model
-            foundation_model_path = model_data.filepath
-            config_parse += f"\nfoundation_model: {foundation_model_path}"
+            with self.inputs.foundation_model.open(mode="rb") as source:
+                with folder.open("modelcopy.model", mode="wb") as target:
```
alinelena, line 180:
```suggestion
                with folder.open("mlff.model", mode="wb") as target:
```

aiida_mlip/calculations/train.py
```diff
@@ -175,9 +176,10 @@ def prepare_for_submission(
...
+                    shutil.copyfileobj(source, target)
+            config_parse += "foundation_model: modelcopy.model"
```
alinelena, line 182:
```suggestion
            config_parse += "foundation_model: mlff.model"
```

aiida_mlip/data/model.py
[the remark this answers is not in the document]
ElliottKasoar: Thanks

I don't think it necessarily needs changing, but it's probably worth understanding where `ModelData_1` gets its name, so we're sure (if you're not already) that it's always the right thing to check against.

aiida_mlip/data/model.py
```diff
@@ -195,13 +155,13 @@ def local_file(
 
     @classmethod
     # pylint: disable=too-many-arguments
-    def download(
+    def from_url(
```
alinelena, line 158: technically probably from_uri is better than from_url

aiida_mlip/calculations/base.py
ElliottKasoar: But is it cleaned up afterwards, since it goes in `An aiida.common.folders.Folder to temporarily write files on disk.`?

Is this what Giovanni was talking about where this is useful for copying the contents stored in a node for remote submission? So it's making the data in the node actually available to be used, essentially? Or is that different?

aiida_mlip/data/model.py
```diff
@@ -35,7 +35,7 @@ class ModelData(SinglefileData):
         Set the file for the node.
     from_local(file, architecture, filename=None):
         Create a ModelData instance from a local file.
-    from_url(url, architecture, filename=None, cache_dir=None, keep_file=False)
+    from_uri(url, architecture, filename=None, cache_dir=None, keep_file=False)
```
oerc0122, line 38: Are you sure you want `URI` here? `URL`s are a subset of `URI`s which provide a resource, but `URI`s also include `URN`s which have no way to reach them and only provide a base-name mapping (from which we couldn't get a file).

If we do want this, need to change all instances (including arguments) to `uri`.

oerc0122 (changes requested):
I think fundamentally the change from `URL` to `URI` is not what we actually want.

aiida_mlip/data/model.py
oerc0122: Why do you think that? We want it from a `URL` `file:///home/...` is a `URL` too, just using a `file` protocol. Likewise `~/my_file` can be considered a `URL` with implicit `file` protocol

aiida_mlip/calculations/base.py
```diff
@@ -311,6 +316,10 @@ def _add_model_to_cmdline(
             # Raise error if model is None (different than model not given as input)
             if self.inputs.model is None:
                 raise ValueError("Model cannot be None")
-            model_path = self.inputs.model.filepath
-        if model_path:
+
+            with self.inputs.model.open(mode="rb") as source:
+                with folder.open("mlff.model", mode="wb") as target:
```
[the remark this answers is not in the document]
oerc0122: However, is there a reason you're using `shutil.copyfileobj` rather than `shutil.copyfile`?

aiida_mlip/data/model.py
```diff
@@ -231,32 +190,44 @@ def download(
...
+        # Looking for ModelData in the whole database
+        for i in qb.iterdict():
+            # If the hash is the same as the new model, but not the creation time
+            # it means that there is already a model that is the same, use that
+            if (
+                "model_hash" in i["ModelData_1"]["attributes"]
+                and i["ModelData_1"]["attributes"]["model_hash"] == model.model_hash
+                and i["ModelData_1"]["attributes"]["architecture"] == model.architecture
+            ):
+                if i["ModelData_1"]["ctime"] != model.ctime:
```
oerc0122, line 220: Is this 2 `if`s, not an extra `and` for a reason?

aiida_mlip/helpers/help_load.py
```diff
@@ -20,17 +20,17 @@ def load_model(
     cache_dir: Optional[Union[str, Path]] = None,
 ) -> ModelData:
     """
-    Load a model from a file path or URL.
+    Load a model from a file path or uri.
```
oerc0122, line 23: It should be noted the `URI` is just as much an acronym as `URL`.

tests/data/test_model.py
```diff
@@ -1,27 +1,27 @@
...
 from pathlib import Path
 
 from aiida_mlip.data.model import ModelData
 
 
 def test_local_file():
-    """Testing that the local file function works"""
+    """Testing that the from_local function works"""
     # Construct a ModelData instance with the local file
     model_path = Path(__file__).parent / "input_files" / "model_local_file.txt"
```
oerc0122, line 11: This seems to be consistently used. Might be best as a module level constant.

aiida_mlip/data/model.py
federicazanca: not really, I can change it. And yeah should have filtering I guess

aiida_mlip/calculations/base.py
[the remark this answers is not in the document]
federicazanca: just copied from aiida's docs example tbh

oerc0122 (approved):

aiida_mlip/data/model.py
```diff
@@ -231,32 +190,38 @@ def download(
...
-        request.urlretrieve(url, file)
+        if keep_file:
+            return model
 
-        if force_download:
-            print(f"filename changed to {file}")
-            return cls.local_file(file=file, architecture=architecture)
+        file.unlink(missing_ok=True)
+
+        qb = QueryBuilder()
```
ElliottKasoar, line 208:
```suggestion
        # Check if the same model was used previously
        qb = QueryBuilder()
```
Did you mean to lose the comments here? It doesn't need to be line by line, but maybe this is enough?

aiida_mlip/calculations/base.py
```diff
@@ -301,6 +303,9 @@ def _add_model_to_cmdline(
         cmd_line : dict
             Dictionary containing the cmd line keys.
 
+        folder : ~aiida.common.folders.Folder
+            An `aiida.common.folders.Folder` to temporarily write files on disk.
```
ElliottKasoar, line 307: I still think this is misleading if it's not temporary, unless I misunderstood our discussion?

federicazanca: no I forgot to change

aiida_mlip/data/model.py
federicazanca: prob cancelled by mistake when I changed the things

ElliottKasoar (approved):
Looks good!

PR #153, opened 2024-07-31
148 add training workchain

federicazanca:
trainign workgraph.
it takes as inputs: folder and the inputs of janus (actually just the janusconfigfile) and quantumespresso's workchain.
it creates the input files for the training
it outputs the optimised structures and the trained model. 

No tests because it will be better to do it when everything is setup in the cloud since the quantumespresso submission is a bit more complex (it might take long times and also we might want to test the ssh connection)


superstar54 (commented):
Thanks for the nice work! I add a few small comments.

PR #161, opened 2024-11-19
Update pre-commit to use ruff

### LINKED ISSUE #156: Update docs
Another pass over the docs would be good. A few things I've already picked up include:

In [getting started](https://stfc.github.io/aiida-mlip/user_guide/get_started.html#getting-started):

- [x] `#pip install -e .[pre-commit,testing] # install extras for more features` needs updating to work with poetry, and probably doesn't need to be in the quick start at all, but rather in the dev guide
- [x] `verdi quicksetup` is deprecated
- [x] The full docstring for `Singlepoint` makes is a bit unreadable
- [ ] `examples/test_submit` has moved/doesn't exist?
- [ ] More specific instructions (here or in the tutorial about adding a computer/code), particularly in the context of having `janus` installed

Developer guide

- [x] Update to match janus-core e.g. correct info for installation via poetry, or otherwise
- [x] We can probably remove publishing and release

Tutorial

- [ ] `Install aiida-mlip, [aiida-core](https://aiida.readthedocs.io/projects/aiida-core/en/latest/intro/get_started.html), [janus-core](https://github.com/stfc/janus-core).` could probably include a link, and/or clarify that installing aiida-mlip should install the requirements anyway

Citation

- [ ] We currently have "Author Name1, Author Name2, Paper title, Jornal Name XXX, YYYY (Year)."
- [ ] We should add a CITATION.cff - see janus-core for example

Other

- [x] I suggest we can probably drop the full repository contents in the README

### LINKED ISSUE #157: Update pre-commit
We should sync changes to janus dev tools, such as replacing pylint, black, etc. with ruff for the pre-commit.

It might be a good opportunity to update the dev guide too.

ElliottKasoar:
Resolves #157

Note: There are newer versions of `ruff` (0.7.4), but these would not allow us to use the latest `janus-core`, so that needs updating first.

I've also added an ignore for `N806` (checks for the use of non-lowercase variable names in functions), as this is throwing errors about the returns from `CalculationFactory` in some, but not all, cases. This should probably be revisited at some point.

Partially resolves #156

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,36 +9,15 @@ repos:
     - id: trailing-whitespace
     - id: check-json
 
-- repo: https://github.com/asottile/pyupgrade
-  rev: v3.15.0
+- repo: https://github.com/astral-sh/ruff-pre-commit
+  # Ruff version.
+  rev: v0.7.4
   hooks:
-  - id: pyupgrade
-    args: ["--py39-plus"]
-
-- repo: https://github.com/PyCQA/isort
-  rev: 5.13.2
-  hooks:
-    - id: isort
-      args: ["--profile=black"]
-
-- repo: https://github.com/psf/black
-  rev: 24.1.1
-  hooks:
-    - id: black
-      args: [--line-length=88]
-
-- repo: local
-  hooks:
-  - id: pylint
-    language: system
-    types: [file, python]
-    name: pylint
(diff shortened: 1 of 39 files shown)


alinelena (approved):

ElliottKasoar:
(Rebased to resolve conflicts with #163)

alinelena (approved):

Issue #166, opened 2024-11-20
Update pytest fixtures


ElliottKasoar:
Currently used fixtures are deprecated, and will be removed with AiiDA version 3, so we should look into using fixtures from `aiida.tools.pytest_fixtures`.

Most fixtures have a direct correspondence to `aiida.manage.tests.pytest_fixtures`, but currently doing this seems to result in the default profile not being set when callling `verdi run`.

This can be fixed along the lines of 

```python
def test_example_md(example_path, aiida_profile, janus_code):
    command = ["verdi", "profile", "set-default", f"{aiida_profile.name}"]
    result = subprocess.run(command, capture_output=True, text=True, check=False)
```

but it ought to be fixable via the fixtures directly.

alinelena:
just be careful how you split the command, passing it like that can be dangerous if not all things are correctly escaped

Issue #164, opened 2024-11-19
Refactor/tidy shared code


ElliottKasoar:
We should do another pass at refactoring code e.g. [converting underscores when defining command line options](https://github.com/stfc/aiida-mlip/pull/162#discussion_r1848348578).

Some tidying up e.g. probably removing [print statements on submission](https://github.com/stfc/aiida-mlip/pull/162#discussion_r1848360511) would be good too.

PR #167, opened 2024-11-20
Fix submission tests

### LINKED ISSUE #165: Fix verdi run tests
Currently the tests that call `verdi run submit_[calc]` return `The code 'janus@localhost' does not exist` in `result.stdout`, suggesting the code is never actually being run.

This can be modified similar to the change for [descriptors](https://github.com/stfc/aiida-mlip/pull/162/commits/dab88e922aab941b67b49622210731ab5952c4ad), but the MD test still fails.

However, this appears to be a more general problem, as `verdi run /home/ek/PSDI/aiida-mlip/examples/calculations/submit_md.py janus@localhost` throws an error:

```
FileNotFoundError: object with path `aiida-traj.xyz` does not exist.
```

whereas the equivalent for singlepoint and geomopt both work.

It may also be worth looking at the migration from `aiida.manage.tests.pytest_fixtures` to `aiida.tools.pytest_fixtures`, as the former is now deprecated. However, swapping these out (which requires changing `aiida_local_code_factory` to `aiida_code_installed`) leads to an `result.stderr`: `Critical: no default profile defined: None`

ElliottKasoar:
Resolves #165

Similar to changes made in #162, this fixes tests for submitting calculations.

As noted in #165, the fixtures we currently use are deprecated, but that can be done separately, as swapping them out seems non-trivial (see #166).

### DIFF
--- a/tests/calculations/test_geomopt.py
+++ b/tests/calculations/test_geomopt.py
@@ -94,12 +94,20 @@ def test_run_opt(model_folder, janus_code):
     assert result["xyz_output"].filename == "aiida-results.xyz"
 
 
-def test_example_opt(example_path):
+def test_example_opt(example_path, janus_code):
     """Test function to run geometry optimization using the example file provided."""
     example_file_path = example_path / "submit_geomopt.py"
-    command = ["verdi", "run", example_file_path, "janus@localhost"]
+    command = [
+        "verdi",
+        "run",
+        example_file_path,
+        f"{janus_code.label}@{janus_code.computer.label}",
+    ]
 
     # Execute the command
     result = subprocess.run(command, capture_output=True, text=True, check=False)
     assert result.stderr == ""
     assert result.returncode == 0
+    assert "results from calculation:" in result.stdout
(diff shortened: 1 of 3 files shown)


oerc0122 (commented):
Looks mostly sensible, though I'm not 100% sure what adding those `assert`s is adding to the validity of the test. Are there more concrete things we might use then just "whether string is printed"?

ElliottKasoar:
> sensible, though I'm not 100% sure what adding those `assert`s is adding to the validity of the test. Are there more concrete things we might use then just "whether string is printed"?

I'm not sure how much we can easily do without relying on calling more `verdi` commands.

If we remove the print statements (#164) we'll probably need to, but at least for now this checks the calculation has actually run, which is the main point of this test, since we inspect the actual outputs in other tests.

alinelena (approved):

oerc0122 (approved):

ElliottKasoar:
I'll rebase this after #168, otherwise the pre-commit will fail

oerc0122 (approved):

PR #169, opened 2024-11-21
Add generic high throughput calculations

ElliottKasoar:
Abstract implementation of #147

I thought it might make sense to make a separate PR, as #147 has some potentially useful, but unused utility functions, among other things, but I've tried to keep the immediately relevant changes.

This takes converts the high throughput geomopt into a more abstract WorkGraph builder, so we can pass any calculation that expects a structure, currently tested for `singlepoint` and `geomopt`.

One change, maybe for a later PR, could be to also allow list of AiiDA output structures, as it would be useful to be able to chain these together e.g.`janus descriptors` -> data reduction -> high throughput singlepoint on reduced set of structures.


examples/tutorials/high_throughput_workgraph.ipynb
```diff
@@ -0,0 +1,338 @@
...
+   "metadata": {},
+   "source": [
+    "## Geometry optimisation"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Similarly, high throughtput geometry optimisation can be performed by changing the `calc`:"
```
ElliottKasoar, line 249:
```suggestion
    "Similarly, high throughput geometry optimisation can be performed by changing the `calc`:"
```

tests/workflows/test_ht.py
```diff
@@ -0,0 +1,81 @@
...
+    wg.run()
+
+    assert wg.state == "FINISHED"
+
+    assert isinstance(wg.process.outputs.final_structures.H2O, SinglefileData)
+    assert isinstance(wg.process.outputs.final_structures.methane, SinglefileData)
+
+
+def test_ht_invalid_path(janus_code, workflow_invalid_folder, model_folder) -> None:
+    """Test high throughput singlepoint calculation."""
```
ElliottKasoar, line 38:
```suggestion
    """Test invalid path for high throughput calculation."""
```

pyproject.toml
```diff
@@ -80,6 +81,9 @@ build-backend = "poetry.core.masonry.api"
 "mlip.train_parser" = "aiida_mlip.parsers.train_parser:TrainParser"
 "mlip.descriptors_parser" = "aiida_mlip.parsers.descriptors_parser:DescriptorsParser"
 
+[tool.poetry.plugins."aiida.workflows"]
+"mlip.ht_wg" = "aiida_mlip.workflows.ht_workgraph:ht_workgraph"
+
```
ElliottKasoar, line 86:
```suggestion
```

examples/tutorials/high_throughput_workgraph.ipynb
```diff
@@ -0,0 +1,338 @@
...
+   "metadata": {
+    "id": "_XkQgKPMtyhf"
+   },
+   "outputs": [],
+   "source": [
+    "inputs = {\n",
+    "    \"model\": model,\n",
+    "    \"metadata\": {\"options\": {\"resources\": {\"num_machines\": 1}}},\n",
+    "    \"code\": code,\n",
+    "    # \"property\": \"energy\",\n",
```
ElliottKasoar, line 114:
```suggestion
    "    \"code\": code,\n",
    "    \"properties\": \"energy\",\n",
```

aiida_mlip/workflows/ht_workgraph.py
```diff
@@ -0,0 +1,126 @@
...
+from aiida.engine import CalcJob, WorkChain
+from aiida_workgraph import WorkGraph, task
+from ase.io import read
+
+from aiida_mlip.helpers.help_load import load_structure
+
+
+@task.graph_builder(outputs=[{"name": "final_structures", "from": "context.structs"}])
+def ht_calc_builder(
+    calc: Union[CalcJob, Callable, WorkChain, WorkGraph],
```
oerc0122, line 15: We haven't switched Aiida-mlip over to `__future__.annotations`, yet? Should at some point to align with Janus

aiida_mlip/workflows/ht_workgraph.py
```diff
@@ -0,0 +1,126 @@
...
+
+    Raises
+    ------
+    FileNotFoundError
+        If `folder` has no valid structure files.
+    """
+    wg = WorkGraph()
+    structure = None
+
+    for child in folder.glob("**/*"):
```
oerc0122, line 55:
```suggestion
    for child in folder.glob("**/*"):
```
`folder.walk()`?

aiida_mlip/workflows/ht_workgraph.py
```diff
@@ -0,0 +1,126 @@
...
+        try:
+            read(child.as_posix())
```
oerc0122, line 57: Why `as_posix`?

aiida_mlip/workflows/ht_workgraph.py
```diff
@@ -0,0 +1,126 @@
...
+        except Exception:
+            continue
+        structure = load_structure(child)
+        calc_inputs[f"{input_struct_key}"] = structure
```
oerc0122, line 61:
```suggestion
        calc_inputs[str(input_struct_key)] = structure
```

aiida_mlip/workflows/ht_workgraph.py
```diff
@@ -0,0 +1,126 @@
...
+        calc_task = wg.add_task(
+            calc,
+            name=f"calc_{child.stem}",
+            **calc_inputs,
+        )
+        calc_task.set_context({f"{final_struct_key}": f"structs.{child.stem}"})
```
oerc0122, line 67:
```suggestion
        calc_task.set_context({str(final_struct_key): f"structs.{child.stem}"})
```

tests/conftest.py
```diff
@@ -240,6 +240,30 @@ def structure_folder(test_folder):
...
+@pytest.fixture
+def workflow_structure_folder(test_folder):
+    """
+    Fixture to provide the path to the example file.
+
+    Returns
+    -------
+        Path: The path to the example file.
+    """
+    return test_folder / "workflows" / "structures"
```
oerc0122, line 252: Do these want to be fixtures or just module level constants?

Possibly want them to be `fixture(scope="session")` to avoid wasting recomputation (incredibly minor, but still)

aiida_mlip/workflows/ht_workgraph.py
ElliottKasoar: Good point, no, I probably should have added that when I added ruff. There's an issue now

aiida_mlip/workflows/ht_workgraph.py
ElliottKasoar: This is only available from Python 3.12 onwards

oerc0122: Weird, so it is. I hate Python versioning. In any case you probably want
```python
    for child in folder.rglob("*"):
```
possibly with
```python
    for child in (for file in folder.rglob("*") if file.is_file()):
```

aiida_mlip/workflows/ht_workgraph.py
ElliottKasoar: Good question. It was in #147, but I can't see any obvious reason (and tests work without it), so I'll remove it.

aiida_mlip/workflows/ht_workgraph.py
ElliottKasoar: Does `rglob` essentially just add the `**/`? I was thinking we may want an option to disable recursive search, so might it make more sense to use glob with/without that?

Agree that checking the file makes sense, thanks

oerc0122: Re `rglob`: Yeah, basically.

tests/conftest.py
ElliottKasoar: They probably could be module level constants, but it's also convenient having these collected. Is there much overhead to a fixture? If not (with the scope added) I think I'd leave it.

PR #177, opened 2025-04-09
Update janus-core and drop Python 3.9

ElliottKasoar:
Updates `janus-core`, which requires dropping Python 3.9, as well as minimum ASE requirement, although this was already used in practice.

We're not affected by the changes janus's output directory, since we specify files by hand, but the changes to how the `geomopt` trajectory is saved needed fixing.

`janus-core` also no longer comes with `mace`, leading to errors unless it is installed.

The output from the tests did not make that at all clear, so something to look into in future is catching and raising this more clearly (although perhaps it's hidden in the AiiDA output files, which are harder to see in tests).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,8 +9,8 @@ jobs:
     timeout-minutes: 30
     strategy:
       matrix:
-        python-version: ['3.9','3.10','3.11','3.12']
-        aiida-version: ['stable']
+        python-version: ["3.10", "3.11", "3.12"]
+        aiida-version: ["stable"]
 
     services:
       postgres:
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -9,8 +9,8 @@ jobs:
     timeout-minutes: 30
     strategy:
       matrix:
-        python-version: ['3.9','3.10','3.11','3.12']
-        aiida-version: ['stable']
+        python-version: ["3.10", "3.11", "3.12"]
```
alinelena, line 12: no 3.13?

ElliottKasoar: `pymatgen` requires `"<3.13,>=3.10"`, so `poetry` complains (it would also be an issue for `torch<2.6`, which the currently released `janus-core` asks for).

PR #178, opened 2025-05-09
Update readme

MM0hsin:
Updated verdi quicksetup to use verdi presto

advise user to setup code to run tutorial

Added a step in development section to install mace

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,6 @@ Create a Python [virtual environment](https://docs.python.org/3/library/venv.htm
 
 ```shell
 pip install aiida-mlip
-verdi presto #Sets up profile
 verdi plugin list aiida.calculations
 ```
 The last command should show a list of AiiDA pre-installed calculations and the aiida-mlip plugin calculations:
@@ -55,16 +54,16 @@ Registered entry points for aiida.calculations:
 * mlip.descriptors
 ```
 
-Aiida-mlip should be ready to run some notebooks at this stage . However, to have [full functionality](https://aiida.readthedocs.io/projects/aiida-core/en/stable/installation/guide_quick.html#quick-install-limitations) we recommend configuring aiida-mlip by creating a profile and setting up a broker.
+Aiida-mlip is installed at this stage . However, to have [full functionality](https://aiida.readthedocs.io/projects/aiida-core/en/stable/installation/guide_quick.html#quick-install-limitations) we recommend configuring aiida-mlip by creating a profile and setting up a broker.
 
 ## AiiDA Configuration
 
-1. Install [RabbitMQ](https://aiida.readthedocs.io/projects/aiida-core/en/stable/installation/guide_complete.html#rabbitmq)
+1. (Optional) Install [RabbitMQ](https://aiida.readthedocs.io/projects/aiida-core/en/stable/installation/guide_complete.html#rabbitmq)
 2. Run:
 ```shell
 verdi presto #Sets up profile and broker for daemon to run
 ```
-`verdi presto` is a quick and simple way to setup the daemon to run some calculations, setting up a profile 'presto'; which configures the computer, broker (i.e. RabbitMQ) and database. [Aiida docs](https://aiida.readthedocs.io/projects/aiida-core/en/stable/installation/guide_complete.html#) go over a more detailed proccess to setup a profile.
+Although setting up a message broker like RabbitMQ is optional, we highly recommend doing so to avoid potential errors and to enable full functionality—such as running the `submit` process. If a broker is detected, the `verdi presto` command offers a quick and easy way to set up the daemon for running calculations. It automatically configures a profile named `presto`, which includes setting up the computer, the database, and the broker (e.g. RabbitMQ). For a more detailed guide on setting up a profile, refer to the [AiiDA documentation](https://aiida.readthedocs.io/projects/aiida-core/en/stable/installation/guide_complete.html#).
 
 ## Usage
 
@@ -92,7 +91,7 @@ a model will be trained using the provided example config file and xyz files (ca
 
 ## Development
 
-We recommend installing uv for dependency management when developing for `aiida-mlip` and setting up PostgreSQL:
+We recommend installing ``uv`` for dependency management when developing for `aiida-mlip`, and setting up PostgreSQL, as this is currently a requirement for testing:
 
 
 1. Install [uv](https://docs.astral.sh/uv/getting-started/installation)
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ The code relies heavily on [janus-core](https://github.com/stfc/janus-core), whi
 
 ```shell
 pip install aiida-mlip
-verdi quicksetup  # better to set up a new profile
+verdi presto  # better to set up a new profile
 verdi plugin list aiida.calculations
 ```
 The last command should show a list of AiiDA pre-installed calculations and the aiida-mlip plugin calculations (mlip.opt, mlip.sp)
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ Registered entry points for aiida.calculations:
 
 ## AiiDA Configuration
 
-Once aiida_mlip is installed, you have to configure aiida-mlip by creating a profile and optionally setting up a broker.
+Once `aiida-mlip` is installed, you have to configure AiiDA by creating a profile to store your data:
 
 1. (Optional) Install [RabbitMQ](https://aiida.readthedocs.io/projects/aiida-core/en/stable/installation/guide_complete.html#rabbitmq)
 2. Run:
--- a/README.md
+++ b/README.md
@@ -32,8 +32,9 @@ machine learning interatomic potentials aiida plugin
 The code relies heavily on [janus-core](https://github.com/stfc/janus-core), which handles mlip calculations using ASE.
 
 
-## Installation
+# Getting Started
 
+## Installation
 Create a Python [virtual environment](https://docs.python.org/3/library/venv.html#creating-virtual-environments) and activate it to install aiida-mlip
 
 ```shell
@@ -52,7 +53,7 @@ Registered entry points for aiida.calculations:
 * mlip.train
 * mlip.descriptors
 ```
-## Prerequisites
+## AiiDA Configuration
 
 Now that aiida-mlip plugin has been installed, we can setup the environment to run calculations:
 
--- a/README.md
+++ b/README.md
@@ -39,9 +39,10 @@ Create a Python [virtual environment](https://docs.python.org/3/library/venv.htm
 
 ```shell
 pip install aiida-mlip
+verdi presto #Sets up profile
 verdi plugin list aiida.calculations
 ```
-The last command should show a list of AiiDA pre-installed calculations and the aiida-mlip plugin calculations (mlip.opt, mlip.sp)
+The last command should show a list of AiiDA pre-installed calculations and the aiida-mlip plugin calculations:
 ```
 Registered entry points for aiida.calculations:
 * core.arithmetic.add
@@ -54,16 +55,14 @@ Registered entry points for aiida.calculations:
 * mlip.descriptors
 ```
 
-Aiida-mlip should be ready to run some notebooks at this stage. However, to have full functionality we reccomend configuring aiida-mlip by creating a profile and setting up a broker.
+Aiida-mlip should be ready to run some notebooks at this stage . However, to have [full functionality](https://aiida.readthedocs.io/projects/aiida-core/en/stable/installation/guide_quick.html#quick-install-limitations) we recommend configuring aiida-mlip by creating a profile and setting up a broker.
 
 ## AiiDA Configuration
 
-Now that aiida-mlip plugin has been installed, we can setup the environment to run calculations:
-
-1. Install RabbitMQ ([ Link ](https://aiida.readthedocs.io/projects/aiida-core/en/stable/installation/guide_complete.html#rabbitmq))
+1. Install [RabbitMQ](https://aiida.readthedocs.io/projects/aiida-core/en/stable/installation/guide_complete.html#rabbitmq)
 2. Run:
 ```shell
-verdi presto #Sets up profile
+verdi presto #Sets up profile and broker for daemon to run
 ```
 `verdi presto` is a quick and simple way to setup the daemon to run some calculations, setting up a profile 'presto'; which configures the computer, broker (i.e. RabbitMQ) and database. [Aiida docs](https://aiida.readthedocs.io/projects/aiida-core/en/stable/installation/guide_complete.html#) go over a more detailed proccess to setup a profile.
 
@@ -93,7 +92,7 @@ a model will be trained using the provided example config file and xyz files (ca
 
 ## Development
 
-We recommend installing uv for dependency management when developing for `aiida-mlip`:
+We recommend installing uv for dependency management when developing for `aiida-mlip` and setting up PostgreSQL:
 
 
 1. Install [uv](https://docs.astral.sh/uv/getting-started/installation)
--- a/README.md
+++ b/README.md
@@ -68,7 +68,9 @@ verdi presto #Sets up profile
 The example folder provides scripts to submit calculations in the calculations folder, and tutorials in jupyter notebook format in the tutorials folder.
 
 A quick demo of how to submit a calculation using the provided example files:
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -84,6 +84,7 @@ We recommend installing uv for dependency management when developing for `aiida-
 git clone https://github.com/stfc/aiida-mlip
 cd aiida-mlip
 uv sync  # Create a virtual environment and install dependencies
+uv sync --extra mace # Install mace seperately
```
ElliottKasoar, line 87: This doesn't need to be a separate command to the one above. I'd maybe also comment that MACE is used in the tests

README.md
```diff
@@ -56,7 +56,7 @@ Registered entry points for aiida.calculations:
 
 The example folder provides scripts to submit calculations in the calculations folder, and tutorials in jupyter notebook format in the tutorials folder.
 
-A quick demo of how to submit a calculation using the provided example files:
+A quick demo of how to submit a calculation using the provided example files (You might need create a [code](https://aiida.readthedocs.io/projects/aiida-core/en/stable/howto/run_codes.html#how-to-create-a-code) for Janus):
```
ElliottKasoar, line 59: Does `verdi presto` set up the computer, or just the profile? If it's just the profile (and maybe either way), I think it's probably best to have a separated comment mentioning the need to set up both the computer and code, rather than having this as an aside.

README.md
```diff
@@ -36,7 +36,7 @@ The code relies heavily on [janus-core](https://github.com/stfc/janus-core), whi
 
 ```shell
 pip install aiida-mlip
-verdi quicksetup  # better to set up a new profile
```
ElliottKasoar, line 39: Maybe it's also worth linking to AiiDA's quick install and complete install, guides somewhere around here, but probably still keep this as a minimal example?

ElliottKasoar (changes requested):

README.md
MMohsinSTFC: `verdi presto` works similar to `quicksetup`, but just sets up rabbitMQ broker and creates a new profile automatically. Its a direct replacement for quicksetup, so we don't need quicksetup

alinelena: documentation or code.
for example in past for profile you could do, of course these are deprecated so we shall understand if it is possible to recover the same behaviour. 
```
verdi setup --config alin.yaml
```

MMohsinSTFC: The recommended way to do that is by using: `verdi profile setup core.psql_dos --config config.yaml` 
So, we can do `verdi presto` for a quick and simple way to automatically setup a default profile for aiida to run in the readme and we can mention > `verdi profile setup core.psql_dos --config config.yaml` in the developer guide.

MMohsinSTFC: [https://aiida.readthedocs.io/projects/aiida-core/en/stable/installation/guide_complete.html#common-options](This) says you have to choose a storage entry point either:
* core.psql_dos
* core.sqlite_dos
* core.sqlite_temp
* core.sqlite_zip

README.md
```diff
@@ -57,6 +57,7 @@ Registered entry points for aiida.calculations:
 The example folder provides scripts to submit calculations in the calculations folder, and tutorials in jupyter notebook format in the tutorials folder.
 
 A quick demo of how to submit a calculation using the provided example files:
+You might need create a [code](https://aiida.readthedocs.io/projects/aiida-core/en/stable/howto/run_codes.html#how-to-create-a-code) for Janus to be recognised as a code by aiida
```
ElliottKasoar, line 60: If you look at your README, this all merges together, so needs some spacing. I'm also not sure it really makes sense in this order.

Perhaps we want a (very brief, mostly made up of links) AiiDA configuration section in between installation and usagem with `verdi presto` etc. also moved in? That can suggest `verdi presto` as the quickest way to get started (+ link), plus mention setting up the code, but then we should probably also link the full set up for more control.

README.md
```diff
@@ -57,6 +57,7 @@ Registered entry points for aiida.calculations:
...
 ```shell
 verdi daemon start     # make sure the daemon is running
```
ElliottKasoar, line 62: This doesn't work with the after just running `verdi presto`, so either need to get it configured before this point, or remove this.

README.md
```diff
@@ -51,12 +52,25 @@ Registered entry points for aiida.calculations:
 * mlip.train
 * mlip.descriptors
 ```
+## Prerequisites
```
ElliottKasoar, line 55: I think this is maybe a bit confusing to have after `Installation`, since it implies these things are required before anything else.

Perhaps it makes sense to have a overall heading of `Getting Started`, and subheadings for `Installation` (same as you have - installing the plugin itself) and then another subheading along the lines of `AiiDA configuration`, which I think should very briefly summarise what needs doing (i.e. configuring a profile/computer/code + optional(?) broker/database), keeping the commands more or less as they are though, just with more context.

README.md
```diff
@@ -51,12 +52,25 @@ Registered entry points for aiida.calculations:
...
+
+Now that aiida-mlip plugin has been installed, we can setup the environment to run calculations:
+
+1. Install RabbitMQ ([ Link ](https://aiida.readthedocs.io/projects/aiida-core/en/stable/installation/guide_complete.html#rabbitmq))
```
ElliottKasoar, line 59: My understanding was still that this is only needed to submit calculations, but it's not strictly necessary to run things? Is that wrong?

MMohsinSTFC: Technically we don't need it for the notebooks, but for the usage section it is important. Users still should setup the daemon, mainly for troubleshooting. Because if there is a problem at least we can be sure they have the bare minimum installed. Also I will change the headings so it should be a bit more clear. But, I do think this step should be included for a first time user so they can run usage steps as a quick way to check if everything is running and setup properly.

ElliottKasoar: Depends on what we say in usage, surely?

Clearly it's a balance, but I'm a bit wary of second-guessing the AiiDA team's decision to separate out running with the daemon from their quickstart with the simpler run commands.

This was also something I encountered/was recommended to me by the devs while working on aiida-workgraph - debugging commands via run is almost always simpler than debugging something you've submitted, and it eliminates a potential source of error (the daemon/RabbitMQ setup).

Why do you say the daemon helps with debugging?

A lot of our docs are written with submission in mind, so we may need to take more of a look at our introductory material alongside these documentation changes, but that's fine too.

README.md
```diff
@@ -32,11 +32,13 @@ machine learning interatomic potentials aiida plugin
...
+
 ## Installation
+Create a Python [virtual environment](https://docs.python.org/3/library/venv.html#creating-virtual-environments) and activate it to install aiida-mlip
 
 ```shell
 pip install aiida-mlip
-verdi quicksetup  # better to set up a new profile
 verdi plugin list aiida.calculations
 ```
 The last command should show a list of AiiDA pre-installed calculations and the aiida-mlip plugin calculations (mlip.opt, mlip.sp)
```
ElliottKasoar, line 44:
```suggestion
The last command should show a list of AiiDA pre-installed calculations and the aiida-mlip plugin calculations:
```
I know this isn't new, but not sure these are necessary when we list them all anyway.

README.md
```diff
@@ -52,11 +54,27 @@ Registered entry points for aiida.calculations:
 * mlip.descriptors
 ```
 
+Aiida-mlip should be ready to run some notebooks at this stage. However, to have full functionality we reccomend configuring aiida-mlip by creating a profile and setting up a broker.
```
ElliottKasoar, line 57: > Aiida-mlip should be ready to run some notebooks at this stage.

Is this true? I think some AiiDA configuration (profile, computer, code (i.e. the minimum that `verdi presto` sets up) is necessary for the notebooks, it's just that we don't need the full config with postgres/RabbitMQ?

I may be wrong though.

README.md
```diff
@@ -77,13 +95,15 @@ a model will be trained using the provided example config file and xyz files (ca
 
 We recommend installing uv for dependency management when developing for `aiida-mlip`:
 
+
 1. Install [uv](https://docs.astral.sh/uv/getting-started/installation)
-2. Install `aiida-mlip` with dependencies in a virtual environment:
+2. Setup [PostgreSQL](https://aiida.readthedocs.io/projects/aiida-core/en/stable/installation/guide_complete.html#core-psql-dos)
+3. Install `aiida-mlip` with dependencies in a virtual environment:
```
ElliottKasoar, line 101: I'd probably either break these up or reword the lead-in sentence, as this essentially suggests this is all related to `uv`.

ElliottKasoar (changes requested):

README.md
ElliottKasoar, line 57:
```suggestion
Aiida-mlip should be ready to run some notebooks at this stage. However, to have full functionality we recommend configuring aiida-mlip by creating a profile and setting up a broker.
```

MMohsinSTFC: Key word being "some", anything that has "load_profile()" wont work. How should I rephrase it?

ElliottKasoar: Do any though? Without having set up anything on the AiiDA side at all, I don't see how it would be storing anything from calculations?

MMohsinSTFC: I got a bit confused, [you mentioned](https://github.com/stfc/aiida-mlip/pull/178#discussion_r2086902048) not needing the daemon but i thought you meant not needing `verdi presto` at all to run the notebooks. So Ill change the line so it reads: 

> Aiida-mlip is installed. However, to have full functionality we recommend configuring aiida-mlip by creating a profile and setting up a broker.

ElliottKasoar: Ah I see, sorry no.

> Aiida-mlip is installed. However, to have full functionality we recommend configuring aiida-mlip by creating a profile and setting up a broker.

I still think we should try to be clearer in terms of distinguishing what is necessary for what functionality.

As I've said, I'm not necessarily against us recommending setting up a broker, but if we do, I think it should be a separate point to the minimal configuration needed to run anything, and I think we need to make it clear what functionality it enables/is unavailable otherwise.

The simpler approach might be to explain it more in terms of functionality than code. E.g. explain that to use AiiDA, we need to set up a profile, computer, code, and optionally the broker (and postgres) with why it's optional. Then we can say this can be done with `verdi presto`, for example, but there's a lot more context behind it?

Then again, that might end up being a bit much, so might be better suited to the long form docs.

docs/source/user_guide/get_started.rst
```diff
@@ -8,55 +8,57 @@ a short example on how to use the plugin.
...
 
 Please refer to the ``janus-core`` `documentation <https://stfc.github.io/janus-core/getting_started/getting_started.html#installation>`_ for further details.
 
 Once ``aiida-mlip`` and the desired MLIP calculators are installed, run::
 
-    verdi quicksetup  # better to set up a new profile
+    verdi presto  # better to set up a new profile
     verdi plugin list aiida.calculations  # should now show your calculation plugins
 
 Then use ``verdi code setup`` with the ``janus`` input plugin
```
ElliottKasoar, line 46:
```suggestion
Then, use ``verdi code setup`` with the ``janus`` input plugin
```

docs/source/user_guide/get_started.rst
```diff
@@ -8,55 +8,57 @@ a short example on how to use the plugin.
...
-to set up an AiiDA code for aiida-mlip.
+to set up an AiiDA code for aiida-mlip. This `notebook <https://github.com/stfc/aiida-mlip/blob/main/examples/tutorials/setup-janus-code.ipynb>`_ may be useful.
```
ElliottKasoar, line 47: Can we link https://aiida.readthedocs.io/projects/aiida-core/en/stable/howto/run_codes.html#how-to-create-a-code here?

docs/source/user_guide/get_started.rst
```diff
@@ -8,55 +8,57 @@ a short example on how to use the plugin.
...
+
 
+.. note:: At this point only a profile has been setup, meaning you may be able to run some notebooks but you will have limited functionality.
+    `Aiida docs <https://aiida.readthedocs.io/projects/aiida-core/en/stable/installation/guide_quick.html#quick-install-limitations>`_
+    explain the limitations of not setting up a broker or PostgreSQL.
```
ElliottKasoar, line 52:
```suggestion
.. note::

    If you have not set up the RabbitMQ message broker, you should be now able to ``run`` processes,
    as demonstrated in the `tutorial notebooks <https://github.com/stfc/aiida-mlip/tree/main/examples/tutorials>`_,
    but cannot ``submit`` processes.
    
    The PostgreSQL database is not configured by default.
    Please refer to the `Aiida documentation <https://aiida.readthedocs.io/projects/aiida-core/en/stable/installation/guide_quick.html#quick-install-limitations>`_
    for more details on the limitations of not setting up a broker or PostgreSQL.
```
Or similar.

Just being a bit more explicit, more rewording is welcome.

README.md
```diff
@@ -32,14 +32,17 @@ machine learning interatomic potentials aiida plugin
...
 
+# Getting Started
+
 ## Installation
+Create a Python [virtual environment](https://docs.python.org/3/library/venv.html#creating-virtual-environments) and activate it to install aiida-mlip
 
 ```shell
 pip install aiida-mlip
-verdi quicksetup  # better to set up a new profile
+verdi presto #Sets up profile
```
ElliottKasoar, line 42:
```suggestion
verdi presto # Sets up profile and localhost computer
```
I think?

ElliottKasoar, line 42: I think I'd still have this in the `AiiDA Configuration` section?

It might be best to merge e.g.

> 1. (Optional) Install RabbitMQ
> 2. verdi presto

With an explanation of why (1) is optional and recommended for simply debugging, but for practical purposes RabbitMQ is recommended to enable all functionality

README.md
```diff
@@ -75,83 +92,23 @@ a model will be trained using the provided example config file and xyz files (ca
 
 ## Development
 
-We recommend installing uv for dependency management when developing for `aiida-mlip`:
+We recommend installing uv for dependency management when developing for `aiida-mlip` and setting up PostgreSQL:
```
ElliottKasoar, line 95:
```suggestion
We recommend installing ``uv`` for dependency management when developing for `aiida-mlip`, and setting up PostgreSQL, as this is currently a requirement for testing:
```

ElliottKasoar (changes requested):
Not your addition, but can we also get rid of

> This page should contain a short guide on what the plugin does and
a short example on how to use the plugin.

(In the user_guide/get_started.rst)

README.md
MMohsinSTFC: What part do you mean to keep in the aiida config section?

ElliottKasoar: Sorry, as in I don't think  `verdi presto` belongs in `Installation`, whereas it does make sense as part of configuring AiIDA.

I would literally just have the pip install and plugin list parts with their text in `Install`, then give all the AiiDA context including the "ready to run" bit etc. in the configuration - it's giving context to that, really.

docs/source/user_guide/get_started.rst
ElliottKasoar, line 1: This generally makes sense, but it also doesn't explain in as much detail as explicitly what's happening in terms of AiiDA configuration etc. It almost might make more sense to completely swap the README and these sections.

Perhaps just a bit more detail here mirroring what we're trying to discuss in the README would be good?

ElliottKasoar (changes requested):

pyproject.toml
```diff
@@ -133,7 +133,7 @@ select = [
     # pycodestyle
     "E", "W",
     # Pyflakes
-    "F",
+    "F", "FA",
```
ElliottKasoar, line 136: I know this is very minor, but it probably ought to be its own PR since it's completely unrelated to the other changes here.

README.md
```diff
@@ -55,16 +54,16 @@ Registered entry points for aiida.calculations:
 * mlip.descriptors
 ```
 
-Aiida-mlip should be ready to run some notebooks at this stage . However, to have [full functionality](https://aiida.readthedocs.io/projects/aiida-core/en/stable/installation/guide_quick.html#quick-install-limitations) we recommend configuring aiida-mlip by creating a profile and setting up a broker.
+Aiida-mlip is installed at this stage . However, to have [full functionality](https://aiida.readthedocs.io/projects/aiida-core/en/stable/installation/guide_quick.html#quick-install-limitations) we recommend configuring aiida-mlip by creating a profile and setting up a broker.
```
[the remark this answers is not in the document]
MMohsinSTFC: I do feel we are getting too into the semantics of it at this point, and it might get a bit too long for a quick start guide. Like if I have said "the program is installed, but to unlock full functionality you should configure it" the implication is that you can use some functionality but it will be limited, meaning that you will eventually have to configure aiida and the extra functionality is explained by the embed link.

ElliottKasoar (approved):

PR #179, opened 2025-05-21
Update pytest fixtures

MM0hsin:
Updated conftest.py with new fixtures

updated tests to load profiles


tests/conftest.py
```diff
@@ -81,14 +81,21 @@ def fixture_localhost(aiida_localhost):
     return localhost
 
 
+@pytest.fixture(scope="module", autouse=True)
+def aiida_profile(aiida_config, aiida_profile_factory):
+    """Create and load a profile."""
+    with aiida_profile_factory(aiida_config) as profile:
+        yield profile
```
ElliottKasoar, line 88: Is this still necessary? It's odd that it would be necessary if we're not doing anything like setting up RabbitMQ

tests/calculations/test_descriptors.py
```diff
@@ -98,9 +98,17 @@ def test_run_descriptors(model_folder, janus_code):
...
     """Test running descriptors calculation using the example file provided."""
     example_file_path = example_path / "submit_descriptors.py"
+    profile = aiida_profile
+    command = [
+        "verdi",
+        "profile",
+        "set-default",
+        profile.name,
+    ]
+    subprocess.run(command, capture_output=True, text=True, check=False)
```
ElliottKasoar, line 111: Given that we need to do this in a bunch of places, maybe it can be refactored (potentially made into a fixture?)

ElliottKasoar (changes requested):

tests/conftest.py
```diff
@@ -81,70 +80,25 @@ def fixture_localhost(aiida_localhost):
...
-    Returns
-    -------
-    `Code`
-        The janus code instance.
+@pytest.fixture(autouse=True)
+def janus_code(aiida_code_installed, aiida_profile):
     """
-    janus_path = shutil.which("janus") or os.environ.get("JANUS_PATH")
-    return aiida_local_code_factory(executable=janus_path, entry_point="mlip.sp")
+    Function-scoped fixture to get the janus Code.
```
ElliottKasoar, line 86:
```suggestion
    Function-scoped fixture to get the janus Code.
```
I'm not against it as such, but why does it make sense to refresh this for every function?

tests/conftest.py
```diff
@@ -81,70 +80,25 @@ def fixture_localhost(aiida_localhost):
...
-
-        Returns
-        -------
-        aiida.orm.nodes.data.code.Code
-            An `InstalledCode` instance.
-        """
-        label = f"test.{entry_point_name}"
-        janus_path = os.environ.get("JANUS_PATH")
-        try:
-            return load_code(label=label)
```
ElliottKasoar, line 138: Is this no longer useful?

tests/calculations/test_descriptors.py
ElliottKasoar, line 1: I'm not against this choice of spacing, but these changes shouldn't be in this PR. if nothing else, if makes it hard to identify what a PR did, and what the history of this file is.

README.md
ElliottKasoar, line 1: Is this supposed to be changed in a few places?

examples/tutorials/wg_simple.ipynb
ElliottKasoar, line 1: I don't think this should be part of this PR

tests/conftest.py
```diff
@@ -2,26 +2,33 @@
 
 from __future__ import annotations
 
+from asyncio import exceptions
```
ElliottKasoar, line 5: Do we not get this from aiida anymore?

tests/conftest.py
```diff
@@ -81,26 +88,19 @@ def fixture_localhost(aiida_localhost):
...
-        A fixture providing a factory for creating local codes.
-
-    Returns
-    -------
-    `Code`
-        The janus code instance.
-    """
+@pytest.fixture
+def janus_code(aiida_code_installed):
+    """Function-scoped fixture to get the janus Code."""
```
ElliottKasoar, line 93: Does this not work as a session scope? There's no obvious reason to need to reload the code each time

tests/conftest.py
```diff
@@ -81,26 +88,19 @@ def fixture_localhost(aiida_localhost):
...
 
+    return aiida_code_installed(
+        label="janus",
+        default_calc_job_plugin="mlip.sp",
+        filepath_executable=janus_path,
+    )
 
-@pytest.fixture
+
+@pytest.fixture(scope="function", autouse=True)
```
ElliottKasoar, line 103: Similar to above, the localhost doesn't change?

ElliottKasoar (changes requested):

ElliottKasoar (commented):
I think this also requires aiida-core>=2.7?

We should update pyproject.toml

tests/conftest.py
MM0hsin: yes we do

ElliottKasoar: But why the swap to `asyncio`?

> actually, we are calling the error on line 137, so we would need this line
> 
> ```
> try:
>      return load_code(label=label)
> except exceptions.NotExistent:
>       return InstalledCode(
> ```

Isn't this what we have?

tests/conftest.py
```diff
@@ -14,14 +14,28 @@
...
     """Automatically clear database in between tests."""
 
 
+@pytest.fixture(scope="session", autouse=True)
+def aiida_profile(aiida_config, aiida_profile_factory):
+    """
+    Session-scoped fixture to create an AiiDA profile.
+
+    Returns
+    -------
```
ElliottKasoar, line 31:
```suggestion
    Yields
    ------
```

PR #180, opened 2025-05-23
Replace opt kwargs

MM0hsin:
replaced instances of opt_kwargs with minimise_kwargs

### DIFF
--- a/aiida_mlip/calculations/geomopt.py
+++ b/aiida_mlip/calculations/geomopt.py
@@ -81,7 +81,7 @@ class GeomOpt(Singlepoint):  # numpydoc ignore=PR01
         )
 
         spec.input(
-            "opt_kwargs",
+            "minimize_kwargs ",
             valid_type=Dict,
             required=False,
             help="Other optimisation keywords",
@@ -121,9 +121,9 @@ class GeomOpt(Singlepoint):  # numpydoc ignore=PR01
             "minimize-kwargs": minimize_kwargs,
             "write-traj": True,
         }
-        if "opt_kwargs" in self.inputs:
-            opt_kwargs = self.inputs.opt_kwargs.get_dict()
-            geom_opt_cmdline["opt-kwargs"] = opt_kwargs
+        if "minimize_kwargs " in self.inputs:
+            minimize_kwargs = self.inputs.minimize_kwargs.get_dict()
+            geom_opt_cmdline["minimize-kwargs"] = minimize_kwargs
         if "opt_cell_fully" in self.inputs:
             geom_opt_cmdline["opt-cell-fully"] = self.inputs.opt_cell_fully.value
         if "opt_cell_lengths" in self.inputs:
--- a/aiida_mlip/calculations/geomopt.py
+++ b/aiida_mlip/calculations/geomopt.py
@@ -81,7 +81,7 @@ class GeomOpt(Singlepoint):  # numpydoc ignore=PR01
         )
 
         spec.input(
-            "minimize_kwargs ",
+            "minimize_kwargs",
             valid_type=Dict,
             required=False,
             help="Other optimisation keywords",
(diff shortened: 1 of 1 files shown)


aiida_mlip/calculations/geomopt.py
```diff
@@ -81,7 +81,7 @@ def define(cls, spec: CalcJobProcessSpec) -> None:
         )
 
         spec.input(
-            "opt_kwargs",
+            "minimize_kwargs",
             valid_type=Dict,
             required=False,
             help="Other optimisation keywords",
```
ElliottKasoar, line 87:
```suggestion
            help="All other GeomOpt keyword arguments",
```
Admittedly, `janus-core` has the same text, but I think it is incorrect.

ElliottKasoar (changes requested):
Thanks for this!

Can you do a `grep` to check for any other places `opt-kwargs`/`opt_kwargs` may be used, even if not in the code. E.g. https://github.com/stfc/aiida-mlip/blob/main/examples/calculations/submit_geomopt.py#L47

I'm quite suspicious that no other changes are necessary.

Could you please add a test that checks we can run `geomopt` calculations with this?

We have an example where `minimize-kwargs` is set, without you needing to change anything, which is quite odd, but that input is not actually `run`: https://github.com/stfc/aiida-mlip/blob/main/tests/calculations/test_geomopt.py#L47

I think we want a new test, more like `test_run_opt`, but with inputs that include `minimize-kwargs`, and check against the output files/logs to ensure it is set correctly. (Have a look at the janus-core tests for examples of testing against it).

ElliottKasoar (changes requested):
> Can you do a grep to check for any other places opt-kwargs/opt_kwargs may be used, even if not in the code. E.g. https://github.com/stfc/aiida-mlip/blob/main/examples/calculations/submit_geomopt.py#L47

I think this still needs updating/checking.

Thanks for adding the new test, it looks good. The cell value seems to be different enough that if you comment out the `filter_kwargs`, it fails, which is nice.

In terms of the `traj_kwargs` `filename`, I think something is going a bit wrong. 

Currently, if we don't have `minimize_kwargs` as an input, the `result` will contain `'remote_folder', 'retrieved', 'log_output', 'std_output', 'xyz_output', 'results_dict', 'traj_file', 'traj_output', 'final_structure'`.

However, if we set `filter_kwargs`, but not `traj_kwargs`, via `minimize_kwargs`, then `result` only contains `'remote_folder', 'retrieved', 'log_output', 'std_output'`, which I think means it's throwing an error.

It's probably related to the default trajectory file getting overwritten by the minimise_kwargs.

Can you look into this please?

MM0hsin:
I have updated: https://github.com/stfc/aiida-mlip/blob/main/examples/calculations/submit_geomopt.py#L47

In terms of traj_kwargs filename, if we want to set something like `"filter_kwargs": {"constant_volume": True},`, we ned a location to add the updated parameters, according to janus-core code: https://github.com/stfc/janus-core/blob/a18c61346939797e65d5122114c68e1f53aa5386/janus_core/calculations/geom_opt.py#L97, if filename is not set, it doesnt create a file. 

If we remove `"traj_kwargs": {"filename": "aiida-traj.xyz"},` and only have `"filter_kwargs": {"constant_volume": True},` the error code mentions:  `[WARNING] output parser returned exit code<305>: Some output files missing or cannot be read`, meaning it is trying to find a file that is not created. If we remove it minimize_kwargs it works because it automatically uses the default values.

ElliottKasoar:
I think this now works as expected.

I've also removed the explicit option to set the trajectory file, similar to what we did in #220 for `precision` once we added `calc_kwargs`, so this is a breaking change.

The main thing we were missing was slightly more careful handling of how `minimize_kwargs` is set, including the default trajectory filename, and ensuring this is known when we define the outputs of the CalcJob, which was a little tricky.

The tests could maybe do a bit more to verify that the settings are being used correctly, but is does at least check that the various options that failed to run at all before are able to run, and I check that the filename is settable in various ways, which was the most likely to go wrong.

oerc0122 (approved):

PR #188, opened 2025-06-24
Update workgraph

MM0hsin:
This PR updates work_graph (0.4.10 ->0.5.3) and aiida-core (2.6.4 -> 2.7.0), fixing conflicting aiida-mlip had with newer implementation


examples/tutorials/high_throughput_workgraph.ipynb
```diff
@@ -153,21 +153,12 @@
     "\n",
     "wg = get_ht_workgraph(\n",
     "    calc=SinglepointCalc,\n",
-    "    folder=Path(\"../../tests/workflows/structures/\"),\n",
+    "    folder=Path(\"/home/mtr46585/aiida-mlip/tests/workflows/structures/\"),\n",
     "    calc_inputs=inputs,\n",
```
ElliottKasoar, line 157: I know you didn't change this, but I think other changes mean `inputs` needs to be changed before this. I get an error now.

The same inputs as singlepoint, without `properties` should be fine.

tests/workflows/test_ht.py
```diff
@@ -54,8 +53,8 @@ def test_ht_invalid_path(janus_code, workflow_invalid_folder, model_folder) -> N
         final_struct_key="xyz_output",
     )
 
-    with pytest.raises(FileNotFoundError):
-        wg.run()
+    wg.run()
+    assert wg.process.exit_code.status == 302
```
ElliottKasoar, line 57: I think we need to be much more specific than this, as this will be the error code for quite a few of the errors we've seen.

If I look at `wg.process.exit_message`, this gives ""WorkGraph finished, but tasks: ['ht_calc'] failed. Thus all their child tasks are skipped."", which is also probably too vague, but a start.

The actual error doesn't seem to be FileNotFoundError anymore though?

```
07/08/2025 10:25:27 AM <51689> plumpy.processes: [ERROR] Error in task ht_calc: TaskSocketNamespace has no attribute 'structs'
Traceback (most recent call last):
  File "/Users/elliottkasoar/Documents/PSDI/aiida-mlip/.venv/lib/python3.12/site-packages/node_graph/socket.py", line 568, in __getattr__
    return self._sockets[name]
           ~~~~~~~~~~~~~^^^^^^
KeyError: 'structs'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/elliottkasoar/Documents/PSDI/aiida-mlip/.venv/lib/python3.12/site-packages/aiida_workgraph/engine/task_manager.py", line 226, in execute_process_task
    process, state = task.execute(
                     ^^^^^^^^^^^^^
  File "/Users/elliottkasoar/Documents/PSDI/aiida-mlip/.venv/lib/python3.12/site-packages/aiida_workgraph/tasks/builtins.py", line 386, in execute
    wg = executor(*args, **kwargs)
         ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/elliottkasoar/Documents/PSDI/aiida-mlip/aiida_mlip/workflows/ht_workgraph.py", line 85, in build_ht_calc
    wg.outputs.final_structures = wg.ctx.structs
                                  ^^^^^^^^^^^^^^
  File "/Users/elliottkasoar/Documents/PSDI/aiida-mlip/.venv/lib/python3.12/site-packages/node_graph/socket.py", line 570, in __getattr__
    raise AttributeError(f"{self.__class__.__name__} has no attribute '{name}'")
AttributeError: TaskSocketNamespace has no attribute 'structs'
```

Do you get the same? If so, I think we want to find a way to handle this better and raise the error more clearly.

ElliottKasoar (changes requested):
Thanks for this!

It's useful to keep wg_simple.ipynb around for now to test things, but I think it probably should be its own PR before this is merged, and so addressing the comments e.g. markdown comments are less urgent, and mostly there for later reference.

tests/workflows/test_ht.py
MM0hsin: Ohyes i do get the same error, but ill update the PR and it will go back to the file not found, i removed a line thinking i didnt need it, even though i tested it before pushing it

PR #191, opened 2025-07-09
Add simple workgraph examples

MM0hsin:
Add a simple work graph example for testing purposes

### DIFF
--- a/examples/tutorials/simple_workgraph.ipynb
+++ b/examples/tutorials/simple_workgraph.ipynb
@@ -30,6 +30,14 @@
   },
   {
    "cell_type": "markdown",
+   "id": "8b8a5703",
+   "metadata": {},
+   "source": [
+    "## Single Point Calculation"
+   ]
+  },
+  {
+   "cell_type": "markdown",
    "id": "9e3e493f",
    "metadata": {},
    "source": [
@@ -377,7 +385,7 @@
    "id": "62cdb8e2",
    "metadata": {},
    "source": [
-    "This time if we want to pass the inputs/outputs directly to the workgraph we can set them manually: "
+    "This time if we want to pass the inputs/outputs directly to the workgraph we can define them:"
    ]
   },
   {
@@ -387,6 +395,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "# As we are passing in the inputs as kwargs we have to pass the inputs directly to the task\n",
     "wg.tasks.geomopt_calc.set(\n",
     "    {**inputs}\n",
     ")"
@@ -399,7 +408,6 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "# We can then set the outputs of the workgraph from the task\n",
     "wg.outputs.results = wg.tasks.geomopt_calc.outputs.results_dict\n",
     "wg.outputs.results_file = wg.tasks.geomopt_calc.outputs.xyz_output"
    ]
--- a/examples/tutorials/simple_workgraph.ipynb
+++ b/examples/tutorials/simple_workgraph.ipynb
@@ -5,7 +5,7 @@
    "id": "ebda8eac",
    "metadata": {},
    "source": [
-    "# Run basic calculations using WorkGraph"
+    "# Run basic calculations using Workgraph"
    ]
   },
   {
@@ -20,20 +20,20 @@
   },
   {
    "cell_type": "markdown",
+   "id": "85abcd21",
+   "metadata": {},
+   "source": [
+    "## Single Point Calculation"
+   ]
+  },
+  {
+   "cell_type": "markdown",
    "id": "6cc3af05",
    "metadata": {},
    "source": [
     "### Setup\n",
     "\n",
     "The initial setup is very similar to the other tutorials, such as `singlepoint.ipynb`, which goes into more detail about what each step is doing"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "id": "8b8a5703",
-   "metadata": {},
-   "source": [
-    "## Single Point Calculation"
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


examples/tutorials/simple_workgraph.ipynb
```diff
@@ -0,0 +1,243 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "517c2066",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "wg.tasks.sp_calc.outputs.xyz_output"
+   ]
+  }
```
ElliottKasoar, line 234: It would be nice to have another section explaining how to set wg-level inputs and outputs (https://aiida-workgraph.readthedocs.io/en/latest/howto/graph_inputs_outputs.html#Example:-Using-graph-level-inputs-and-outputs), since particularly setting the outputs is useful for more complex workflows.

Perhaps have a geomopt example with both for some variety (making it's clear it's not essential, just illustrative).

ElliottKasoar (commented):
Looks good generally, thanks! I've left a few comments about how I'd like us to document this.

The final comment about another example could potentially be done later.

examples/tutorials/simple_workgraph.ipynb
MM0hsin: oh sorry, I meant having a another section on how to set graph level inputs and outputs. Having them in one section will, in my opinion, make it look cleaner as we are setting up the whole workgraph. We can use comments in the section to explain how the graph level input/outputs are set.

Also, yes ill add another section for geomopt calculation

examples/tutorials/simple_workgraph.ipynb
```diff
@@ -0,0 +1,507 @@
...
+    "    name=\"geomopt_calc\",\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "62cdb8e2",
+   "metadata": {},
+   "source": [
+    "This time if we want to pass the inputs/otuputs directly to the workgraph we can set them manually: "
```
ElliottKasoar, line 374:
```suggestion
    "This time if we want to pass the inputs/outputs directly to the workgraph we can set them manually: "
```

examples/tutorials/simple_workgraph.ipynb
```diff
@@ -0,0 +1,507 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "32e9a6a5",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# We can then set the outputs of the workgraph from the task\n",
```
ElliottKasoar, line 396: Do we need this as well as the markdown?

examples/tutorials/simple_workgraph.ipynb
```diff
@@ -0,0 +1,507 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "1ba44195",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "wg.tasks.geomopt_calc.set(\n",
```
ElliottKasoar, line 384: Is this the same as setting `wg.inputs`? It's probably worth generally being slightly more explicit that it's not just "directly" setting things, but defining things on a workgraph level rather than a task level.

examples/tutorials/simple_workgraph.ipynb
```diff
@@ -0,0 +1,507 @@
...
+   "source": [
+    "# Geometry Optimisation"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "ddf8a780",
+   "metadata": {},
+   "source": [
+    "The calculations can be setup with WorkGraph very similarily in most cases, in this example we will setup a geometry optimisation calculation. First we setup the the inputs and load the calculation"
```
ElliottKasoar, line 308:
```suggestion
    "The calculations can be setup with WorkGraph very similarly in most cases. In this example we will setup a geometry optimisation calculation. First we setup the the inputs and load the calculation"
```

examples/tutorials/simple_workgraph.ipynb
```diff
@@ -0,0 +1,507 @@
...
+   "source": [
+    "wg.tasks.sp_calc.outputs.results_dict.value.get_dict()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "7609565d",
+   "metadata": {},
+   "source": [
+    "We can also use verdi to interact with the calcaltion and see the output (you may need to click `view scollable element` if you are running the command in a notebook to see the mose recent tasks)"
```
ElliottKasoar, line 264:
```suggestion
    "We can also use verdi to interact with the calcaltion and see the output (you may need to click `view scrollable element` if you are running the command in a notebook to see the mose recent tasks)"
```

examples/tutorials/simple_workgraph.ipynb
```diff
@@ -0,0 +1,507 @@
...
+    "\n",
+    "generate_node_graph(wg.pk)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "becbcddf",
+   "metadata": {},
+   "source": [
+    "We can then check the output to ensure we are getting the correct output. Like checking the xyz_output which should return a `SinglefileData` object"
```
ElliottKasoar, line 228:
```suggestion
    "We can then check the output to ensure we are getting the correct output. Here, we check the `xyz_output`, which should return a `SinglefileData` object"
```

MM0hsin (commented):
Address PR comments

examples/tutorials/simple_workgraph.ipynb
ElliottKasoar: Ok, I see what you mean now.

This looks more like showing an alternative way to set the task-level inputs (rather than doing it when the task is created as we do for single point), not wg-level inputs, but it's still useful to see, and I think the explanation is enough.

Issue #189, opened 2025-07-08
Create map for nested workgraphs


ElliottKasoar:
> Also separately, because we have nested workgraph we should ideally be creating a map: https://aiida-workgraph.readthedocs.io/en/latest/howto/map.html (could be a new issue) because the whole point of the workgraph gui is to show the tasks and calculations going on, and see the internal calculations like calc_H2O and calc_methane

_Originally posted by @MM0hsin in https://github.com/stfc/aiida-mlip/pull/188#discussion_r2192622932_

PR #204, opened 2025-07-31
Update WorkGraph v0.6.0

MM0hsin:
Bump aiida-workgraph 0.6.0 and update tests

### DIFF
--- a/aiida_mlip/workflows/ht_workgraph.py
+++ b/aiida_mlip/workflows/ht_workgraph.py
@@ -20,7 +20,7 @@ def build_ht_calc(
     input_struct_key: str = "struct",
     final_struct_key: str = "final_structure",
     recursive: bool = True,
-):
+) -> WorkGraph:
     """
     Build high throughput calculation WorkGraph.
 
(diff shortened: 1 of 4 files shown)


MM0hsin:
> So do we no longer need the equivalent to graph_builder for dynamic inputs then, because of the way the context manager system now works?

Yes, because the inputs keys (code, model, etc...) stay more or less the same. 

We can also close: [#189](https://github.com/stfc/aiida-mlip/issues/189) as its no longer needed

pyproject.toml
```diff
@@ -25,7 +25,7 @@ dependencies = [
     "ase<4.0,>=3.24",
     "voluptuous<1,>=0.15.2",
     "janus-core<0.8,>=0.7.5",
-    "aiida-workgraph<0.6, >=0.5.3",
+    "aiida-workgraph<0.6.9, >=0.6.0",
```
ElliottKasoar, line 28:
```suggestion
    "aiida-workgraph<0.7, >=0.6.0",
```
0.6.10 is a perfectly valid release

aiida_mlip/workflows/ht_workgraph.py
```diff
@@ -55,94 +54,35 @@ def build_ht_calc(
...
+            calc_inputs[input_struct_key] = structure
+
+            calc_task = wg.add_task(
+                calc,
+                name=f"calc_{file.stem}",
+                **calc_inputs,
+            )
+            structs[file.stem] = getattr(calc_task.outputs, final_struct_key)
+
+        wg.outputs.final_structure = structs
```
ElliottKasoar, line 81: It's pretty minor, but why is `final_structure` singular now?

In general we want variable names to give context, but particularly for an output designed specifically to be interacted with by users, the distinction between `structure` and `structures` is useful.

ElliottKasoar (changes requested):

ElliottKasoar (changes requested):
There's one more `final_strucuture` in the notebook, but otherwise I think this should be good to merge.

ElliottKasoar (approved):

PR #210, opened 2025-08-04
Fix high throughput geomopt example

ElliottKasoar:
The input to `geomopt` included `properties` from the single point example, so it raises errors when the workgraph is run.


MM0hsin (approved):
Looks good!

PR #202, opened 2025-07-28
Add geomopt, descriptors and fine tuning

MM0hsin:

PR #213, opened 2025-08-12
Create notebook to run WG with two calculations on a structure

MM0hsin:


examples/tutorials/sample_split.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+def write_samples(frames, inds, f_s):
+    """Write selected frames to a file."""
+    for i in inds:
+        write(f_s, frames[i], write_info=True, append=True)
+
+
+@task.calcfunction()
+def process_and_split_data(**inputs):
```
ElliottKasoar, line 40: A return type here and as part of the docstring would help here

examples/tutorials/sample_split.py
```diff
@@ -0,0 +1,176 @@
...
+    valid_file = Path(f"{prefix}valid.xyz")
+    test_file = Path(f"{prefix}test.xyz")
+
+    if not append_mode:
+        _ = [p.unlink(missing_ok=True) for p in [train_file, valid_file, test_file]]
+
+    print(f"create files: {train_file=}, {valid_file=} and {test_file=}")
+
+    stats = {}
+    for i, f in enumerate(a):
```
ElliottKasoar, line 85: Something more meaningful like `struct` would be much clearer than `f`.

examples/tutorials/descriptors_filter.ipynb
```diff
@@ -0,0 +1,287 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from aiida.orm import Str, Float, Bool, Int\n",
+    "from ase.io import read\n",
+    "from aiida_workgraph import WorkGraph\n",
+    "from aiida.orm import StructureData\n",
+    "from sample_split import process_and_split_data\n",
+    "\n",
+    "initail_structure = \"../../examples/tutorials/structures/lj-traj.xyz\"\n",
```
ElliottKasoar, line 161: Looks like a typo in the variable name, but also I don't think you should need `../../examples/tutorials`, when this is in `examples/tutorials`?

examples/tutorials/descriptors_filter.ipynb
```diff
@@ -0,0 +1,287 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "86036116",
+   "metadata": {},
+   "source": [
+    "In the loop we can call each structure; \n",
+    "* Run geomopt calculation \n",
+    "* Pass the output file into and run descriptors calculation \n",
```
ElliottKasoar, line 144: We should be clearer what the output (file) is that we want to pass onwards, as there are several outputs.

examples/tutorials/descriptors_filter.ipynb
```diff
@@ -0,0 +1,287 @@
...
+   "source": [
+    "## Aim"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "35ecbb39",
+   "metadata": {},
+   "source": [
+    "This notebook shows how we can run multiple calculations on a given structure"
```
ElliottKasoar, line 24: These two descriptions are a bit different, and neither is particularly clear.

The focus should be on the types of task/output types we're connecting.

ElliottKasoar (changes requested):

pyproject.toml
```diff
@@ -32,6 +32,14 @@ dependencies = [
 mace = [
     "janus-core[mace]"
 ]
+qe = [
+    "aiida-quantumespresso<5,>=4.12.1",
+]
```
ElliottKasoar, line 37: This isn't used in this PR, is it?

examples/tutorials/descriptors_filter.ipynb
```diff
@@ -0,0 +1,304 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "d2f463f6",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "file = open(\"test.xyz\")\n",
+    "print(file.read())"
```
ElliottKasoar, line 279: Do we not need to get this from the outputs?

Either way, we really want to load this with ase, not just the raw contents.

examples/tutorials/workgraphs/descriptors_filter.ipynb
```diff
@@ -0,0 +1,345 @@
...
+   "id": "1795dd41",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from aiida.orm import Str, Float, Bool\n",
+    "inputs = {\n",
+    "    \"code\": janus_code,\n",
+    "    \"model\": model,\n",
+    "    \"arch\": Str(model.architecture),\n",
+    "    \"precision\": Str(\"float64\"),\n",
```
ElliottKasoar, line 101:
```suggestion
```
I don't think this is (or should be) used?

examples/tutorials/workgraphs/descriptors_filter.ipynb
```diff
@@ -0,0 +1,345 @@
...
+    "num_structs = len(read(initail_structure, index=\":\"))\n",
+    "\n",
+    "with WorkGraph(\"Calculation Workgraph\") as wg:\n",
+    "    wg.inputs = inputs\n",
+    "    final_structures = {}\n",
+    "\n",
+    "    for i in range(num_structs):\n",
+    "        structure = StructureData(ase=read(initail_structure, index=i))\n",
+    "\n",
+    "        geomopt_calc = wg.add_task(\n",
```
ElliottKasoar, line 180: For these inputs, does it maybe make more sense to have different input dicts for each task?

This feels quite error prone if you have set manually things from the shared `inputs`. We can always have a shared dict and task-specific dicts if you find lots of duplication.

examples/tutorials/workgraphs/descriptors_filter.ipynb
```diff
@@ -0,0 +1,345 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "0e4c78c9",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from aiida.plugins import CalculationFactory\n",
+    "\n",
+    "geomoptCalc = CalculationFactory(\"mlip.opt\")\n",
+    "descriptorsCalc = CalculationFactory(\"mlip.descriptors\")\n"
```
ElliottKasoar, line 128:
```suggestion
    "descriptorsCalc = CalculationFactory(\"mlip.descriptors\")"
```
Very minor, to be a little neater

examples/tutorials/workgraphs/descriptors_filter.ipynb
```diff
@@ -0,0 +1,345 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "5acf0b46",
+   "metadata": {},
+   "source": [
+    "### Setup\n",
+    "\n",
+    "The initial setup is very similar to the other tutorials, such as `singlepoint.ipynb` which goes into more detail about what each step is doing.\n",
```
ElliottKasoar, line 34:
```suggestion
    "The initial setup is very similar to the other tutorials, such as `singlepoint.ipynb`, which goes into more detail about what each step is doing.\n",
```
Very minor

ElliottKasoar (commented):
Generally looking really nice!

As discussed in meetings with @MM0hsin, the plan is for most tidying up of sample_split.py to be left for #202, so it doesn't need to be perfect here.

I've left a few minor tidying up suggestions, but the main thing I want to consider is the inputs to the tasks, to make it easier to track what is input where.

examples/tutorials/workgraphs/descriptors_filter.ipynb
MM0hsin: Not a massive advantage, other than to see all the inputs that are going in when you create/build the workgraph. The input nodes arn't always visible unless you pass in a kwargs dictionary or if theres a socket going in.

PR #220, opened 2025-08-29
Add calc kwargs input

### LINKED ISSUE #216: How to set additional flags for MACECalculators
Hi I'm a new user of `aiida-mlip` and I can run the tutorial: https://github.com/stfc/aiida-mlip/blob/main/examples/tutorials/singlepoint.ipynb

Just wondering how can I set the additional arguments for `MACECalculator` (in ASE) such as `enable_cueq`? I tried to put them in the `inputs` dictionary but I got following error:
```
Error occurred validating port 'inputs': Unexpected ports {'enable_cueq': True}, for a non dynamic namespace
```

ElliottKasoar:
Resolves #216

Replaces `precision` base input for `calc_kwargs` dictionary input, which should allow any additional key word arguments to be passed to `janus-core`.

A few things to tidy up:

- [x] Update example CLIs with calc_kwargs input (and test at least one) 
- [x] Add note on use in docs
- [x] Add example of use in at least one tutorial

Note: Uses #219 for base, due to setting `model` option, so this will require rebasing.


harveydevereux (approved):
Looks alright to me

PR #225, opened 2025-09-09
Make command line args conversion a function

### LINKED ISSUE #126: make common pattern a function
Given this is such a common pattern to all of these would it be best to turn this into a function?

_Originally posted by @oerc0122 in https://github.com/stfc/aiida-mlip/pull/123#discussion_r1601603611_

oerc0122:
Simple new function to reduce code duplication and simplify code.

Fixes #126

### DIFF
--- a/aiida_mlip/helpers/converters.py
+++ b/aiida_mlip/helpers/converters.py
@@ -123,7 +123,7 @@ def kwarg_to_param(params: dict[str, Any]) -> list[str]:
     >>> kwarg_to_param({"name": "Geoff", "key": True})
     ['--name', 'Geoff', '--key']
     >>> kwarg_to_param({"value": 6, "falsey": False})
-    ['--value', '6']
+    ['--value', '6', '--no-falsey']
     """
     cmdline_params = []
 


alinelena (dismissed):
looks good can we get the ci green?

alinelena (dismissed):

aiida_mlip/helpers/converters.py
```diff
@@ -99,3 +100,41 @@ def convert_to_nodes(dictionary: dict, convert_all: bool = False) -> dict:
...
+    """
+    cmdline_params = []
+
+    for key, val in params.items():
+        key = key.replace("_", "-")
+        match val:
+            case bool() if val:
+                cmdline_params.append(f"--{key}")
+            case bool():
+                ...
```
ElliottKasoar, line 136: I may just not be thinking this through enough, but doesn't this assume the default is `False`?

If the default is `True`, we need to set the `--no-key` flag, which I don't think this does?

oerc0122: Are there any parameters like this? Are all parameters compatible with `--no-key`?

ElliottKasoar: I'm not certain there are any currently settable in aiida-mlip, but there are certainly are few in janus e.g. `janus eos` defaults to `--minimize`, and everything currently defaults to `--tracker` (which we probably need to make sure we can disable).

If it's a boolean, it would always be `--no-key`, I believe. We just define the `key`, and it automatically creates the two flags. We probably could overwrite this in theory, but I don't think we do.

PR #227, opened 2025-09-09
Update example submission script

### LINKED ISSUE #224: Submit using config out of date
There seem to be quite a few issues with this script (https://github.com/stfc/aiida-mlip/blob/main/examples/calculations/submit_using_config.py), including:

- A `model` is not defined
- The loaded structure path means it only works when running from the examples directory
- The `JanusConfigfile` path is relative, so raises an error

This code is also in the calculations documentation, so that will need updating similarly.

We should probably add it to the tests, as we have for the other submit examples.

ElliottKasoar:
Resolves #224

I think the original idea was that `model` would be set in the config file, which is why it was not previously included in the example, but given that we already set the structure in a way that cannot be overwritten, I think also setting the model is fine as an example, as it still shows the use of the config file?

Alternatively, we could add some example `model` path to the config file (with a workaround needed for testing purposes), which may be preferable?

### DIFF
--- a/docs/source/user_guide/calculations.rst
+++ b/docs/source/user_guide/calculations.rst
@@ -123,7 +123,14 @@ They will be converted to AiiDA data types by the script itself.
 
     verdi run submit_singlepoint.py "janus@localhost" --structure "path/to/structure" --model "path/to/model" --device "cpu"
 
-The ``submit_using_config.py`` script can be used to facilitate submission using a config file.
+The ``submit_using_config.py`` script provides an example of submission using a config file.
+
+.. note::
+
+    The structure and model are hard-coded into ``submit_using_config.py`` to avoid
+    issues with relative paths. These should be modified, or removed and set through
+    the configuration file, for your structure and model of interest.
+
 
 Geometry Optimisation calculation
 ---------------------------------


examples/calculations/submit_using_config.py
```diff
@@ -2,32 +2,45 @@
...
 from aiida.orm import load_code
 from aiida.plugins import CalculationFactory
 
+import aiida_mlip
 from aiida_mlip.data.config import JanusConfigfile
-from aiida_mlip.helpers.help_load import load_structure
+from aiida_mlip.helpers.help_load import load_model, load_structure
+
+# __file__ changes depending on where verdi run is called
+DATA_PATH = Path(aiida_mlip.__file__).parent.parent / "tests" / "calculations"
```
oerc0122, line 16: Is this relevant/cleaner? https://docs.python.org/3/library/pkgutil.html#pkgutil.get_data

ElliottKasoar: It looks like it's useful for loading the contents of a file quite nicely, but I don't think it gives us a nicer way to get the path, which we need to construct the AiiDA data types?

oerc0122: True, the data aren't in the package.

oerc0122 (approved):

PR #238, opened 2025-10-02
Fix parsing error with yaml file

MM0hsin:
Update train.py to add attribute to new line, to fix issues when parsing training attributes from aiida-mlip to janus


ElliottKasoar (changes requested):
Thanks for this! A few things that would be useful for you to add/confirm:

- A little bit of explanation about what the root of the bug actually was. I/others can infer from the fix, but having it spelled out in a bit more detail is helpful to confirm exactly what was going wrong
- A test that fails before this change is applied, but passes now
- This change suggests the error only occurred with fine tuning. Is this the case?

Issue #229, opened 2025-09-10
Add pressure option for geomopt


ElliottKasoar:
`janus-core` prevents pressure being set through `minimize_kwargs`, instead requiring it to be set explicitly, and so even with #152 resolved, we are unable to run geometry optimisation at a specific pressure.

Issue #192, opened 2025-07-09
Add and test tutorials in documentation


ElliottKasoar:
We should consider moving the current tutorials into the documentation, and testing them. See https://github.com/stfc/janus-core/pull/399.

PR #239, opened 2025-10-03
#229 pressure for geomopt

JessieGould:
Pressure added to :

- geomopt.py
- test_geomopt.py
- submit_geomopt.py

### DIFF
--- a/aiida_mlip/calculations/geomopt.py
+++ b/aiida_mlip/calculations/geomopt.py
@@ -121,15 +121,7 @@ class GeomOpt(Singlepoint):  # numpydoc ignore=PR01
             "minimize-kwargs": minimize_kwargs,
             "write-traj": True,
         }
-        # if "opt_cell_fully" in self.inputs:
-        #     geom_opt_cmdline["opt-cell-fully"] = self.inputs.opt_cell_fully.value
-        # if "opt_cell_lengths" in self.inputs:
-        #     geom_opt_cmdline["opt-cell-lengths"] = self.inputs.opt_cell_lengths.value
-        # if "fmax" in self.inputs:
-        #     geom_opt_cmdline["fmax"] = self.inputs.fmax.value
-        # if "steps" in self.inputs:
-        #     geom_opt_cmdline["steps"] = self.inputs.steps.value
-        # if "pressure" in self.inputs:
+
         for key in ("opt_cell_fully", "opt_cell_lengths", "fmax", "steps", "pressure"):
             if key in self.inputs:
                 geom_opt_cmdline[key] = getattr(self.inputs, key).value


examples/calculations/submit_geomopt.py
```diff
@@ -108,6 +109,9 @@ def geomopt(params: dict) -> None:
 @click.option(
     "--steps", default=1000, type=int, help="Maximum number of optimisation steps."
 )
+@click.option(
+    "--pressure", default=5.0, type=float, help="Pressure to use for the calculation."
```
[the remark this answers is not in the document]
JessieGould: this is now done

aiida_mlip/calculations/geomopt.py
```diff
@@ -82,6 +82,13 @@ def define(cls, spec: CalcJobProcessSpec) -> None:
             help="All other keyword arguments to pass to geometry optimizer",
         )
 
+        spec.input(
+            "pressure",
+            valid_type=Float,
+            required=False,
+            help="External pressure (float)",
```
ElliottKasoar, line 89: I don't think we need to say it's a float, since that's labelled by the `valid_type` already, but something closer to the CLI help (in GPa) is probably useful.

ElliottKasoar (changes requested):
This looks great so far, thanks!

I think the biggest thing left is to make sure we can set `pressure` and that it changes the outputs in new tests.

One short would could be  a test similar to the `test_geomopt` function, where we just check that a non-zero pressure input results in `--pressure` and the value being passed to the command line.

I would also include a test where we set the `--pressure` `click` input, run the calculation, and check that the output values change with respect to 0 pressure (which hopefully we can see from one of the existing tests, such as `test_run_geomopt`).

Note: you (should) need to make sure one of the `opt_cell` booleans is True for `pressure` to change the output.

Another nice-to-have would be adding to the geomopt tutorial, and to the list of inputs in the docs (https://stfc.github.io/aiida-mlip/user_guide/tutorial.html#usage), to reflect the new option.

examples/tutorials/calculations/geometry-optimisation.ipynb
ElliottKasoar, line 1: This is one of the downsides of Jupyter notebooks (motivating exploring alternatives in #192): they don't play very nicely with git.

When you add notebooks, can you clear the results, so it would show as `execution_count": null,`, and try to add only the relevant code changes.

If you use `git add -p`, you can control exactly which lines are added from a file, and then you can discard changes such as `version": "3.12.3"`, which leads to more apparent changes than are relevant.

JessieGould: I saw that too, I restored to main and then added the single line and re-pushed. It looks good again to me, hopefully the same your end. I also ran the geometry-optimisation.ipynb and realised it wasn't recognising the pressure, so you'll see that I have added it to the base.py

aiida_mlip/calculations/base.py
```diff
@@ -190,6 +190,13 @@ def define(cls, spec: CalcJobProcessSpec) -> None:
             help="Name of the log output file",
         )
 
+        spec.input(
+            "pressure",
+            valid_type=Float,
+            required=False,
+            help="Scalar pressure when optimizing cell geometry, in GPa.",
+        )
+
```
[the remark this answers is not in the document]
JessieGould: Thanks for highlighting. I've just checked and it works without it, I'll remove it now

tests/calculations/test_pressure.py
```diff
@@ -0,0 +1,216 @@
...
+            "struct": StructureData(ase=bulk("NaCl", "rocksalt", 5.63)),
+            "model": ModelData.from_local(model_file, architecture="mace"),
+            "device": Str("cpu"),
+            "opt_cell_fully": Bool(True),
+            "fmax": Float(0.1),
+            "steps": Int(1000),
+            "pressure": Float(pressure),
+            "minimize_kwargs": Dict(
+                {"traj_kwargs": {"filename": f"traj-{pressure}GPa.xyz"}}
+            ),
```
ElliottKasoar: We probably don't need this. All it's really doing is checking we change the filename, which we do elsewhere.

oerc0122 (dismissed):
Looks good if it works and other concerns are addressed. Just one comment that may not be this PR.

PR #240, opened 2025-10-06
Rename CLI options

### LINKED ISSUE #110: Rename CLI options
Currently CLI options use underscores to separate words, which differs from the default for click/typer option naming (see `Parameters` [here](https://click.palletsprojects.com/en/8.1.x/api/)) (and so also from `janus` currently), and more generally from [GNU conventions](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html).

We certainly want to be consistent with janus, and I think sticking to the standard would be the best approach.

Note: Using hyphens was an issue for reading the config for janus, but that's now [fixed](https://github.com/stfc/janus-core/pull/113). A similar fix may be needed for reading aiida inputs, but anything passed to janus should be fine.

While changing option names, we may want to consider if we want to unify some of the shared options e.g. `arch`/`architecture`

_Originally posted by @ElliottKasoar in https://github.com/stfc/aiida-mlip/pull/104#discussion_r1555982989_

MM0hsin:
Edit submit_[calc].py files, where CLI args use `_` to seperate words in args. Replace them with `-`.

Minor fix - submit_md_using_config.py, changed config path to a general absolute path.

Resolves #110

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ A quick demo of how to submit a calculation using the provided example files:
 verdi daemon start     # make sure the daemon is running
 cd examples/calculations
 verdi run submit_singlepoint.py "janus@localhost" --struct "path/to/structure" --architecture mace --model "/path/to/model"    # run singlepoint calculation
-verdi run submit_geomopt.py "janus@localhost" --struct "path/to/structure" --model "path/to/model" --steps 5 --opt_cell_fully True # run geometry optimisation
+verdi run submit_geomopt.py "janus@localhost" --struct "path/to/structure" --model "path/to/model" --steps 5 --opt-cell-fully True # run geometry optimisation
 verdi run submit_md.py "janus@localhost" --struct "path/to/structure" --model "path/to/model" --ensemble "nve" --md-dict-str "{'temp':300,'steps':4,'traj-every':3,'stats-every':1}" # run molecular dynamics
 
 verdi process list -a  # check record of calculation
--- a/tests/calculations/test_geomopt.py
+++ b/tests/calculations/test_geomopt.py
@@ -372,7 +372,7 @@ def test_pressure_verdi_run(example_path, janus_code, tmp_path):
         str(structure_file),
         "--pressure",
         "5.0",
-        "--opt_cell_fully",
+        "--opt-cell-fully",
         "True",
     ]
     result_5_pressure = subprocess.run(


examples/calculations/submit_md_using_config.py
```diff
@@ -18,8 +20,12 @@
 
 # All the other paramenters we want them from the config file
 # We want to pass it as a AiiDA data type for the provenance
+
+
 config = JanusConfigfile(
-    "/home/federica/aiida-mlip/tests/calculations/configs/config_janus_md.yaml"
+    Path("~/aiida-mlip/tests/calculations/configs/config_janus_md.yml")
```
ElliottKasoar, line 26: I think this probably ought to be the (expanded/resolved/absolute) relative path, since this assumes almost as much about the user's file structure as the original.

ElliottKasoar (changes requested):
Generally looks good, thanks!

Can you add at least one of the geomopt inputs to `test_geomopt.py`, to check they can be set correctly still?

Can you also check the documentation for any references to these old inputs? I'm pretty sure there's at least one reference to `--md_dict_str`.

examples/calculations/submit_md_using_config.py
ElliottKasoar: No, you can get the path to this file, and use a path relative to that to get the absolute path to the config, completely independent to where the base `aiida-mlip` directory is.

examples/calculations/submit_md_using_config.py
```diff
@@ -18,8 +20,12 @@
 
 # All the other paramenters we want them from the config file
 # We want to pass it as a AiiDA data type for the provenance
+
+
 config = JanusConfigfile(
-    "/home/federica/aiida-mlip/tests/calculations/configs/config_janus_md.yaml"
+    Path("../../../tests/calculations/configs/config_janus_md.yml")
+    .expanduser()
```
ElliottKasoar: This isn't doing anything?

examples/calculations/submit_md_using_config.py
```diff
@@ -18,8 +20,10 @@
 
 # All the other paramenters we want them from the config file
 # We want to pass it as a AiiDA data type for the provenance
+
+
 config = JanusConfigfile(
-    "/home/federica/aiida-mlip/tests/calculations/configs/config_janus_md.yaml"
+    Path("../../../tests/calculations/configs/config_janus_md.yml").resolve()
```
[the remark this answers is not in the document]
MM0hsin: created #257

PR #250, opened 2025-11-05
192 add and test tutorials in documentation

### LINKED ISSUE #192: Add and test tutorials in documentation

JessieGould:
closes #192


JessieGould:
"We should consider moving the current tutorials into the documentation, ..."
---- the calculations ones are now in docs with nbsphinx (ie not the workgraph ones ) 


"...and testing them. See https://github.com/stfc/janus-core/pull/399."
---- troubleshooting / updating
---- tidying


" This may also include removing/reworking the current Usage and Tutorial sections. "
---- I have drafted a welcome page for docs  / readme_tutorials for devs
---- think about logical order 
---- include basic.ipynb from aiida

JessieGould (commented):
consider removing pandoc from toml? ie is it necessary

docs/source/conf.py
```diff
@@ -247,3 +248,5 @@ def run_apidoc(_):
 
 def setup(app):
     app.connect("builder-inited", run_apidoc)
+
+nbsphinx_allow_errors = True
```
ElliottKasoar: Why do we need this?

.github/workflows/ci.yml
```diff
@@ -79,8 +99,27 @@ jobs:
           version: "0.6.14"
           python-version: "3.12"
 
+      - name: Install pandoc
+        uses: pandoc/actions/setup@v1
+
       - name: Install dependencies
-        run: uv sync
+        run: |
+          uv sync --group docs
```
ElliottKasoar, line 107: We can leave this but just to note `docs` is a default group, so I don't think this should be needed?

docs/source/user_guide/notebooks/config_sp.yaml
```diff
@@ -0,0 +1,5 @@
+arch: mace_mp
+calc-kwargs:
+    calc_kwargs:
+      dispersion: True
```
ElliottKasoar, line 4: Is this correct? The input to the CalcJob would just be `"calc_kwargs": Dict({"dispersion": True}),`, no nesting needed.

docs/source/user_guide/get_started.rst
```diff
@@ -80,3 +80,6 @@ These are the available calculations
    * Train
 
 For more details on the calculations, please refer to the `calculations section <https://stfc.github.io/aiida-mlip/user_guide/calculations.html>`_.
+
+If you would like a refresher on how to use AiiDA in general, please refer to the `AiiDA user guide <https://aiida.readthedocs.io/projects/aiida-core/en/stable/user_guide/index.html>`_.
+An excellent introduction to Aiida Basics can be found in the `AiiDA Basics tutorial <https://aiida-tutorials.readthedocs.io/en/latest/sections/getting_started/basics.html>`_.
```
ElliottKasoar, line 85:
```suggestion
An excellent introduction to AiiDA Basics can be found in the `AiiDA Basics tutorial <https://aiida-tutorials.readthedocs.io/en/latest/sections/getting_started/basics.html>`_.
```

ElliottKasoar (commented):
Looking good so far!

We need to update the REAME, as this is no longer true:

> The [examples](https://github.com/stfc/aiida-mlip/tree/main/examples) folder provides scripts to submit calculations in the calculations folder, and tutorials in jupyter notebook format in the tutorials folder.

I'd probably also add links to these in `docs/source/user_guide/tutorial.rst`.

Note the first line of that is wrong since I think it should be a geom opt:

> This tutorial guides you through running a single-point calculation using the aiida-mlip package.

ElliottKasoar:
I may be missing something obvious, but are these tutorials actually being built and added to the docs?

I don't think they're added to any of the indexes (e.g. see https://github.com/stfc/janus-core/blob/main/docs/source/tutorials/python/index.rst), and there are no changes e.g.  `nbsphinx_execute=never` to the Makefile for `make html`, so I don't think there's any way to not build them if there were being built?

PR #255, opened 2025-11-20
187 add equation of state workgraph

MM0hsin:
Add EOS workgraph

Get Scales > QE and split into test, train and valid file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,8 +82,8 @@ A quick demo of how to submit a calculation using the provided example files:
 verdi daemon start     # make sure the daemon is running
 cd examples/calculations
 verdi run submit_singlepoint.py "janus@localhost" --struct "path/to/structure" --architecture mace --model "/path/to/model"    # run singlepoint calculation
-verdi run submit_geomopt.py "janus@localhost" --struct "path/to/structure" --model "path/to/model" --steps 5 --opt_cell_fully True # run geometry optimisation
-verdi run submit_md.py "janus@localhost" --struct "path/to/structure" --model "path/to/model" --ensemble "nve" --md_dict_str "{'temp':300,'steps':4,'traj-every':3,'stats-every':1}" # run molecular dynamics
+verdi run submit_geomopt.py "janus@localhost" --struct "path/to/structure" --model "path/to/model" --steps 5 --opt-cell-fully True # run geometry optimisation
+verdi run submit_md.py "janus@localhost" --struct "path/to/structure" --model "path/to/model" --ensemble "nve" --md-dict-str "{'temp':300,'steps':4,'traj-every':3,'stats-every':1}" # run molecular dynamics
 
 verdi process list -a  # check record of calculation
 ```
--- a/docs/source/user_guide/calculations.rst
+++ b/docs/source/user_guide/calculations.rst
@@ -201,4 +201,4 @@ They will be converted to AiiDA data types by the script itself.
 
 .. code-block:: python
 
-    verdi run submit_md.py "janus@localhost" --structure "path/to/structure" --model "path/to/model" --ensemble "nve" --md_dict_str "{'temp':300,'steps':4,'traj-every':3,'stats-every':1}"
+    verdi run submit_md.py "janus@localhost" --structure "path/to/structure" --model "path/to/model" --ensemble "nve" --md-dict-str "{'temp':300,'steps':4,'traj-every':3,'stats-every':1}"
--- a/examples/calculations/submit_geomopt.py
+++ b/examples/calculations/submit_geomopt.py
@@ -95,13 +95,13 @@ def geomopt(params: dict) -> None:
 )
 @click.option("--fmax", default=0.1, type=float, help="Maximum force for convergence.")
 @click.option(
-    "--opt_cell_lengths",
+    "--opt-cell-lengths",
     default=False,
     type=bool,
     help="Optimise cell vectors, as well as atomic positions.",
 )
 @click.option(
-    "--opt_cell_fully",
+    "--opt-cell-fully",
     default=False,
     type=bool,
     help="Fully optimise the cell vectors, angles, and atomic positions.",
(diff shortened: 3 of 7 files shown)


examples/tutorials/workgraphs/scales_workgraph.ipynb
```diff
@@ -0,0 +1,469 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "e5c9a2fb",
+   "metadata": {},
+   "source": [
+    "## Aim\n",
+    "\n",
+    "This notebook demonstrates how we can connect and execute tasks in a workflow. In this example; we get an initail structure, generate scaled structures, run `Quantum Espresso` and then split structures into `test`, `train` and `valid` files."
```
oerc0122:
```suggestion
    "This notebook demonstrates how we can connect and execute tasks in a workflow. In this example; we get an initial structure, generate scaled structures, run `Quantum Espresso` and then split structures into `test`, `train` and `valid` files."
```

examples/tutorials/workgraphs/scales_workgraph.ipynb
```diff
@@ -0,0 +1,469 @@
...
+    "\n",
+    "For this tutorial we will assume you have: \n",
+    "<ul>\n",
+    "        <li>An AiiDA profile setup</li>\n",
+    "        <li>An external computer setup in AiiDA with a quantum espresso code</li>\n",
+    "                <ul>\n",
+    "                        <li>A tutorial can be found in <a href=\"../aiida_setup/setup-external-computer.ipynb\"><code>../aiida_setup/setup-external-computer.ipynb</code></a></li>\n",
+    "                </ul>\n",
+    "        <li>The <code>aiida-quantumespresso</code>, <code>aiida-pseudo</code> and <code>fpsample</code> extra dependancies installed</li>\n",
+    "        <li>Pseudopotentails SSSP installed</li>\n",
```
oerc0122:
```suggestion
    "        <li>Pseudopotentials SSSP installed</li>\n",
```

examples/tutorials/workgraphs/scales_workgraph.ipynb
```diff
@@ -0,0 +1,469 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "b96d1ae5",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "test_file = wg.tasks.create_train_files.outputs.test_file.value\n",
+    "train_file = wg.tasks.create_train_files.outputs.train_file.value\n",
+    "valid_file = wg.tasks.create_train_files.outputs.valid_file.value"
```
oerc0122: I think this block and the one below would be better as a series of `dict`s so you can iterate nicely.


```suggestion
    "data_files = {\n",
    "    \"test\": wg.tasks.create_train_files.outputs.test_file.value\n",
    "    \"train\": wg.tasks.create_train_files.outputs.train_file.value\n",
    "    \"valid\": wg.tasks.create_train_files.outputs.valid_file.value\n",
    "}\n"
```
Then you can do:
```python
energies = {
    key: np.array([struct.info['qe_energy'] 
                   for struct in iread(path, index=':')])
    for key, file in data_files.items()
}
lens = np.cumsum([len(x) for x in energies.values()])
axes = {key: np.arange(n) for key, n in zip(energies, lens)}

fig, ax = plt.subplots(figsize=(12, 5))

for (key, x_axis), energy in zip(axes.items(), energies.values(), strict=True):
    ax.bar(x_axis, energy, label=key.capitalize())

y_min = min(map(np.min, energies.values()))
y_max = max(map(np.max, energies.values()))
```

And reduce a lot of the code duplication.

examples/tutorials/workgraphs/scales_workgraph.ipynb
```diff
@@ -0,0 +1,469 @@
...
+   "source": [
+    "wg.run()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "f8a6b9ee",
+   "metadata": {},
+   "source": [
+    "We can can use the outputs to viusualise the data, in this example we will print a simple bar chart of the energies:"
```
ElliottKasoar: It would also be nice to actually have the EoS plot here (see examples linked elsewhere). All you need is the energies (from the DFT) and the volumes (which you can get from the structures if this is not saved).

There's probably no time right now, but it would also be cool to then do an `eos` via janus with the fine-tuned model, and see how that compares.

ElliottKasoar (commented):
How does this compare in terms of implementation to similar AiiDA example workflows, e.g. https://aiida-tutorials.readthedocs.io/en/latest/sections/writing_workflows/realworld.html and https://workgraph-collections.readthedocs.io/en/stable/cp2k/eos.html?

examples/tutorials/workgraphs/eos_fine_tuning.ipynb
```diff
@@ -0,0 +1,469 @@
...
+    "    training_split = {\n",
+    "        \"test\":structures_stack[:i1],\n",
+    "        \"train\":structures_stack[i1:i2],\n",
+    "        \"valid\":structures_stack[i2:]\n",
+    "    }\n",
+    "\n",
+    "    files = {}\n",
+    "\n",
+    "    for split, split_structures in training_split.items():\n",
+    "        tmpfile = tempfile.NamedTemporaryFile(suffix=f\"{split}.extxyz\")\n",
```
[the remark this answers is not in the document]
oerc0122: I don't see why that would mean no `with`?

MM0hsin:
@ElliottKasoar 
> How does this compare in terms of implementation to similar AiiDA example workflows


I was not aware workgraph-collections existed before this. But main difference is we dont use `ase`  `EquationOfState` directly to calculate the energies, instead we generate the scaled structures and run `Quantum Espresso`

PR #258, opened 2025-11-28
Updates authors

MM0hsin:


pyproject.toml
```diff
@@ -7,6 +7,7 @@ authors = [
     { name = "Elliott Kasoar" },
     { name = "Jacob Wilkins" },
     { name = "Alin M. Elena" },
+    { name = "Muhammad Mohsin" },
```
ElliottKasoar:
```suggestion
    { name = "Elliott Kasoar" },
    { name = "Jacob Wilkins" },
    { name = "Muhammad Mohsin" },
    { name = "Alin M. Elena" },
```
Generally we have Federica first, Alin last, and otherwise alphabetical

ElliottKasoar (commented):
There is also a [citation file](https://github.com/stfc/aiida-mlip/blob/main/CITATION.cff) that we should update

ElliottKasoar (dismissed):

ElliottKasoar:
Note: replaced the docs authors with an automated system, as we do in janus-core and elsewhere, and I've reordered the names slightly to be alphabetical by surname in between Federica and Alin, again as we do in janus-core.

CITATION.cff
```diff
@@ -5,12 +5,17 @@ type: software
...
   - family-names: Zanca
     given-names: Federica
+  - family-names: Gould
+    given-names: Jessica
+    orcid: "https://orcid.org/0009-0009-3570-1820"
   - family-names: Kasoar
     given-names: Elliott
+  - family-names: Mohsin
+    given-names: Muhammad
     orcid: "https://orcid.org/0009-0005-2015-9478"
```
oerc0122: This is @ElliottKasoar 's orcid. Double check

ElliottKasoar: Ah that's my bad copy and paste when I shuffled things around I think, sorry!

oerc0122 (approved):

Issue #262, opened 2025-12-12
JSON error when loading model in calculation


MM0hsin:
When running singlepoint.ipynb the cell which loads the model:

```
from aiida_mlip.data.model import ModelData
uri = "https://github.com/stfc/janus-core/raw/main/tests/models/mace_mp_small.model"
model = ModelData.from_uri(uri, architecture="mace_mp", cache_dir="mlips")
```
Throws the error:
```
File /usr/lib/python3.12/json/encoder.py:200, in JSONEncoder.encode(self, o)
    196         return encode_basestring(o)
    197 # This doesn't pass the iterator directly to ''.join() because the
    198 # exceptions aren't as detailed.  The list call should be roughly
    199 # equivalent to the PySequence_Fast that ''.join() would do.
--> [200](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/usr/lib/python3.12/json/encoder.py:200) chunks = self.iterencode(o, _one_shot=True)
    201 if not isinstance(chunks, (list, tuple)):
    202     chunks = list(chunks)

File /usr/lib/python3.12/json/encoder.py:258, in JSONEncoder.iterencode(self, o, _one_shot)
    253 else:
    254     _iterencode = _make_iterencode(
    255         markers, self.default, _encoder, self.indent, floatstr,
    256         self.key_separator, self.item_separator, self.sort_keys,
    257         self.skipkeys, _one_shot)
--> [258](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/usr/lib/python3.12/json/encoder.py:258) return _iterencode(o, 0)

File /usr/lib/python3.12/json/encoder.py:180, in JSONEncoder.default(self, o)
    161 def default(self, o):
    162     """Implement this method in a subclass such that it returns
    163     a serializable object for ``o``, or calls the base implementation
    164     (to raise a ``TypeError``).
   (...)    178 
    179     """
--> [180](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/usr/lib/python3.12/json/encoder.py:180)     raise TypeError(f'Object of type {o.__class__.__name__} '
    181                     f'is not JSON serializable')

TypeError: Object of type datetime is not JSON serializable
```
### Steps to replicate:

On a fresh machine the ran the commands: 
```
   79  curl -LsSf https://astral.sh/uv/install.sh | sh
   80  which uv
   81  cd ./aiida-mlip/
[...]

Issue #261, opened 2025-12-11
Update janus-core


ElliottKasoar:
Currently blocked since `aiida-core` depends on `click>=8.1.0,<8.2.dev0`, but `janus-core` depends on click>8.2.1. This should be resolved once `aiida-core` 2.8 is released: https://github.com/orgs/aiidateam/projects/18, as the relevant PR has been merged.

ajjackson:
Looks like 8.2 is now viable 👍

ajjackson:
But this may be unnecessary: see #269

ElliottKasoar:
> Looks like 8.2 is now viable 👍

Yes, but unfortunately this is likely blocked by a serialisation error we get with newer versions: https://github.com/stfc/aiida-mlip/issues/262.

On top of this, both `aiida-quantumespresso` and `aiida-pseudo` have pins on click.

`aiida-pseudo` pinned it as a [temporary measure](https://github.com/aiidateam/aiida-pseudo/commit/a36ded647683235d777d1230b4cd504ee96d42fc) to cover for the aiida-core's errors, so hopefully we can ask them to bump it, and I think `aiida-quantumespresso` just haven't bumped it in a while.

ajjackson:
It seems a generally bad strategy for plugins to strictly pin any non-niche dependencies. I appreciate that the mess with click is a bit exceptional and revolves around one particular problematic release. But the real headache arises because different projects have dealt with this inconsistently: aiida-pseudo and aiida-core pinning _below_ and janus-core pinning _above_. Does `janus-core` really need a feature that is in click 8.2.1 and not in 8.1.*? Could it instead have something like `click~=8.1,!=8.2.1`?

`aiida-quantumespresso` has `click~=8.0`, that's not a hard pin and allows any click 8.* version.

(I still think the situation is improved by dropping `janus-core` from requirements, #269 )

ElliottKasoar:
> It seems a generally bad strategy for plugins to strictly pin any non-niche dependencies. I appreciate that the mess with click is a bit exceptional and revolves around one particular problematic release. But the real headache arises because different projects have dealt with this inconsistently: aiida-pseudo and aiida-core pinning _below_ and janus-core pinning _above_. Does `janus-core` really need a feature that is in click 8.2.1 and not in 8.1.*? Could it instead have something like `click~=8.1,!=8.2.1`?

Firstly, I don't really agree that `click<9,>=8.2.1` is strict my most standards.

[...]

PR #268, opened 2026-03-19
Update aiida-core and dependencies versions

alinelena:


ElliottKasoar:
See comments here: https://github.com/stfc/aiida-mlip/issues/261.

#262 is the most fundamental issue, but we probably need to ask/open PRs on the two aiida plugins to bump their dependencies too.

oerc0122 (dismissed):

ElliottKasoar:
We need to bump workgraph for this, I think.

If I remember correctly, there are still conflicts with janus too due to aiida-pseudo (both our optional dependency, and via aiida-qe, since they fixed and released the same issue in aiida-qe), which should be resolved once they do a release. I can ping them and see if there's anything stopping this.

I'm also not sure we want to pin to aiida unless we need to. They're generally pretty careful about keeping their API stable, and in the case where they broke things, they patched it pretty quickly

PR #270, opened 2026-03-26
Remove janus_core as dependency, instead require external setup

oerc0122:


ajjackson:
I think you could clear the RabbitMQ complaints out of test output by changing the workflow to use 3.8.14 instead of "latest".

I can't find reference to RabbitMQ version constraints in the AiiDA docs any more, so it might be that the latest `aiida-core` is happier with an up-to-date RabbitMQ.

ElliottKasoar:
> I think you could clear the RabbitMQ complaints out of test output by changing the workflow to use 3.8.14 instead of "latest".
> 
> I can't find reference to RabbitMQ version constraints in the AiiDA docs any more, so it might be that the latest `aiida-core` is happier with an up-to-date RabbitMQ.

We're unfortunately not currently compatible with the last few releases of `aiida-core`, but it also seems to still have the same compatibility issues with the default configuration, the note is just a bit hidden away: https://aiida.readthedocs.io/projects/aiida-core/en/latest/installation/troubleshooting.html#rabbitmq-incompatibility.

ajjackson:
The default RabbitMQ timeout of 30 mins should be fine for CI, so another option is to disable the warning with `verdi config set warnings.rabbitmq_version False`

PR #274, opened 2026-07-25
Upgrade to aiida-core 2.7.4

agoscinski:
aiida-core v2.7.4 includes upstream bug fix aiidateam/aiida-core#7486 resolving issue stfc/aiida-mlip#262. The datetime serialization in the Querybuilder was not handled correctly.

CI passes on my fork see https://github.com/agoscinski/aiida-mlip/actions/runs/30156792551

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ readme = "README.md"
 keywords = ["aiida", "plugin"]
 
 dependencies = [
-    "aiida-core==2.7.1",
+    "aiida-core~=2.7.4",
     "ase<4.0,>=3.24",
     "voluptuous<1,>=0.15.2",
     "janus-core<0.9,>=0.8.3",


ElliottKasoar (approved):
Hi @agoscinski, thanks so much for looking into this!

~Sorry(?) the solution was quite underwhelming in the end, I really thought I'd encountered similar issues with later versions of `aiida-core`. Perhaps I'm just conflating other conflicts that meant we couldn't upgrade this far previously.~ Ahh sorry I completely missed that it was an upstream fix, thank you for that!

From a quick test, things all still look good with 2.8.1, so once a new `aiida-workgraph` is released, we should be able to be up to date with everything.

agoscinski:
Yes sorry, maybe I should be more elaborate. As you probably figured out, the bug exists also in v2.8.0. I also fixed it in v2.8.1

PR #275, opened 2026-08-26
Allow list properties in singlepoint

oerc0122:
Allow passing list of parameters to builder

### DIFF
--- a/tests/calculations/test_singlepoint.py
+++ b/tests/calculations/test_singlepoint.py
@@ -6,10 +6,11 @@ import subprocess
 
 from aiida.common import InputValidationError, datastructures
 from aiida.engine import run
-from aiida.orm import Dict, Str, StructureData
+from aiida.orm import Dict, List, Str, StructureData
 from aiida.plugins import CalculationFactory
 from ase.build import bulk
 from ase.io import write
+import numpy as np
 import pytest
 
 from aiida_mlip.data.config import JanusConfigfile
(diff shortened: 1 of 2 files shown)


trdurrant:
Hi @ElliottKasoar @alinelena It would be great to get this merged in as it would be very helpful to the ALC-UX project

ElliottKasoar:
This looks good, but can you add a test for multiple properties? I'd suggest testing the Hessian from mace, since other properties are often calculated even if they're not requested.

trdurrant:
@oerc0122

ElliottKasoar:
@trdurrant how urgent is this? I think @oerc0122 is away at least today, but I can quickly add the test and Alin can review if needed?

tests/calculations/test_singlepoint.py
```diff
@@ -28,6 +28,7 @@ def test_singlepoint(fixture_sandbox, generate_calc_job, janus_code, model_folde
         "struct": StructureData(ase=bulk("NaCl", "rocksalt", 5.63)),
         "model": ModelData.from_local(model_file, architecture="mace"),
         "device": Str("cpu"),
+        "properties": List(["energy", "forces"]),
```
ElliottKasoar: I would be inclined to leave this one as it was, since I think it was the minimum set of inputs for a single point, and it so it serves as a test for the default set of command line inputs. Then the `cmdline_params` can be checked in the other test, which would test exactly the same as this.

tests/calculations/test_singlepoint.py
```diff
@@ -151,6 +156,37 @@ def test_run_sp(model_folder, janus_code):
...
+        "properties": ["energy", "hessian"],
+    }
+
+    SinglepointCalc = CalculationFactory("mlip.sp")
+    result = run(SinglepointCalc, **inputs)
+    assert "results_dict" in result
+    obtained_res = result["results_dict"].get_dict()
+    assert "xyz_output" in result
+    assert "mace_stress" not in obtained_res["info"]
+    assert "mace_forces" not in obtained_res["info"]
```
ElliottKasoar, line 203: Forces wouldn't be stored in `info` (I think it may be in `obtained_res` directly), so this doesn't test what it should be testing really.

I think neither stress nor forces are present, so maybe this a sensible thing to check, but I'm still surprised.

The results xyz file produced by `janus singlepoint --arch mace_mp --struct tests/data/NaCl.cif --properties hessian --properties energy` (using the janus test data) includes both forces and stress, so do you know why they're not present?

(I assume it's something simple I'm just missing)

tests/calculations/test_singlepoint.py
```diff
@@ -151,6 +152,68 @@ def test_run_sp(model_folder, janus_code):
...
+    assert sorted(calc_info.codes_info[0].cmdline_params) == sorted(cmdline_params)
+
+    SinglepointCalc = CalculationFactory(entry_point_name)
+    result = run(SinglepointCalc, **inputs)
+
+    assert "results_dict" in result
+    obtained_res = result["results_dict"].get_dict()
+    assert "xyz_output" in result
+    assert "mace_stress" not in obtained_res["info"]
+    assert "mace_forces" not in obtained_res["info"]
```
ElliottKasoar, line 203:
```suggestion
    assert "mace_forces" not in obtained_res
```
I think the current form would always be true, even with `mace_forces`

tests/calculations/test_singlepoint.py
ElliottKasoar: I think stress and force are saved in the extxyz in this case, but these are not converted as part of the `todict` we do as part of parsing this, since they're automatically added as singlepoint calculator results, whereas our custom-labeled copies (which are only done for `properties`) are not.

See code suggestion for what I think the force check should be

tests/calculations/test_singlepoint.py
alinelena: arent't forces in arrays rather than info object?

ElliottKasoar: If we were reading it directly from `Atoms`, yes, but when the `todict` conversion seems to drop this. If I print `obtained_res` with all of the properties requested, it looks like:

```
{
    'numbers': [11, 17],
    'positions': [[0.0, 0.0, 0.0], [2.815, 0.0, 0.0]],
    'masses': [22.98976928, 35.45],
    'mace_forces': [[0.0, 0.0, -0.0], [-0.0, -0.0, 0.0]],
    'cell': [[0.0, 2.815, 2.815], [2.815, 0.0, 2.815], [2.815, 2.815, 0.0]],
    'pbc': [True, True, True],
    'info': {
        'units': {
            'energy': 'eV', 'forces': 'ev/Ang', 'stress': 'ev/Ang^3', 'hessian': 'ev/Ang^2'
        },
        'mace_hessian': [[[1.6544188146878, -6.9389542182217e-18, 7.8437107851185e-18], [-1.6544188146878, 9.5410867715164e-18, -5.2138497639941e-18]], [[-1.3871787596087e-17, 1.6544188146878, -1.5958084231669e-18], [1.2134906971593e-17, -1.6544188146878, -2.0208690258226e-18]], [[-1.6318782958569e-17, 6.2895179973864e-18, 1.6544188146878], [1.6377849771122e-17, -7.0413667614921e-18, -1.6544188146878]], [[-1.6544188146878, 6.9389542182217e-18, -7.8437107851185e-18], [1.6544188146878, -9.5410867715164e-18, 5.2138497639941e-18]], [[1.3871787596087e-17, -1.6544188146878, 1.5958084231669e-18], [-1.2134906971593e-17, 1.6544188146878, 2.0208690258226e-18]], [[1.6318782958569e-17, -6.2895179973864e-18, -1.6544188146878], [-1.6377849771122e-17, 7.0413667614921e-18, 1.6544188146878]]],
        'emissions': 3.7920524666753e-06,
        'model': 'mlff.model',
        'arch': 'mace',
        'mace_energy': -6.7575203839729,
        'mace_stress': [-0.005816546985101, -0.005816546985101, -0.005816546985101, -4.9521739247094e-19, -8.4089397009791e-20, 9.763081935595e-20],
        'system_name': 'aiida'
    }
}
```

ElliottKasoar: I've added predicting forces too, so there's no ambiguity for this. I left stress excluded, and check it doesn't appear in either case just to be completely sure

alinelena (approved):

